PR #5, opened 2016-05-02
Fix SL

mpkocher:
Still some package naming inconsistencies that needs to be addressed.


mpkocher:
The failures are because of the cert files aren't checked in. Need to address this issue.

PR #16, opened 2016-05-03
started smrt-server-tools project, with initial ServiceAccessLayer API and get-status replacement

natechols:
This builds and get-status is runnable, but it exits with code 1 and I can't figure out why.

### DIFF
--- a/smrt-server-tools/build.sbt
+++ b/smrt-server-tools/build.sbt
@@ -1,0 +1,8 @@
+
+initialCommands in (Test, console) := """ammonite.repl.Main.run("")"""
+
+packSettings
+
+packMain := Map(
+  "get-status" -> "com.pacbio.secondary.smrttools.tools.GetStatusApp"
+)
--- a/smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/client/ServiceAccessLayer.scala
+++ b/smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/client/ServiceAccessLayer.scala
@@ -1,0 +1,52 @@
+// derived from PrimaryClient.scala in PAWS
+package com.pacbio.secondary.smrttools.client
+
+import java.net.URL
+
+import akka.actor.ActorSystem
+import spray.client.pipelining._
+import spray.http._
+import spray.httpx.SprayJsonSupport
+import spray.json.DefaultJsonProtocol
+
+import scala.concurrent.Future
+
+object ClientModels {
+  // the xfer service isn't upto date with all the fields
+  case class ServiceStatus(id: String, message: String)
+}
+
+// XXX should this use PacBioJsonProtocol instead?
+trait ServicesClientJsonProtocol extends DefaultJsonProtocol {
+  import ClientModels._
+ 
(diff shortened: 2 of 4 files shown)


smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/tools/GetStatus.scala
```diff
@@ -0,0 +1,94 @@
...
+object GetStatusRunner {
+
+  def apply (c: GetStatusConfig): Int = {
+    val startedAt = DateTime.now()
+
+    implicit val actorSystem = ActorSystem("get-status")
+    val url = new URL(s"http://${c.host}:${c.port}")
+    //logger.debug(s"url: ${url}")
+    val exitCode =
+      Try {
```
mpkocher: For debugging it might useful to log the error.  
```
val tx = Try {}
val exitcode = tr match {
case Success(_) => 0 
case Failure(ex) => 
logger.error("failed with ${ex.getMessage}") 
1
```

natechols: I'm confused, how would this fit in the the code currently inside the Try{} block?

natechols:
Oddly, when I get rid of the Try{} altogether, it works exactly as intended.  So I'm just going to check that in since I find it much easier to deal with uncaught exceptions for now anyway.

natechols:
I don't understand these test failures at all, but they don't look like mine.

mpkocher:
They're not related. I believe they're due to indeterminism, or issues tests running in parallel.

smrt-server-tools/build.sbt
```diff
@@ -0,0 +1,8 @@
+
+initialCommands in (Test, console) := """ammonite.repl.Main.run("")"""
+
+packSettings
+
+packMain := Map(
+  "get-status" -> "com.pacbio.secondary.smrttools.tools.GetStatusApp"
```
mpkocher: Because this tool is general, perhaps it should be renamed to `get-smrt-server-status` (or something along those lines)?

smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/client/ServiceAccessLayer.scala
```diff
@@ -0,0 +1,52 @@
...
+import spray.client.pipelining._
+import spray.http._
+import spray.httpx.SprayJsonSupport
+import spray.json.DefaultJsonProtocol
+
+import scala.concurrent.Future
+
+object ClientModels {
+  // the xfer service isn't upto date with all the fields
+  case class ServiceStatus(id: String, message: String)
```
mpkocher: Eventually, I think we should remove this in favor of com.pacbio.common.models.ServiceStatus.

natechols: Yup, already done in my local copy.

mpkocher: Great.

PR #21, opened 2016-05-04
starting to flesh out the service access layer and tools

natechols:
With a functioning stub for 'pbservice'.

### DIFF
--- a/smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/client/ServiceAccessLayer.scala
+++ b/smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/client/ServiceAccessLayer.scala
@@ -3,6 +3,7 @@ package com.pacbio.secondary.smrttools.client
 
 import com.pacbio.secondary.analysis.constants.{GlobalConstants, FileTypes}
 import com.pacbio.secondary.smrtlink.models._
+import com.pacbio.secondary.analysis.jobs.JobModels._
 import com.pacbio.common.models._
 
 import akka.actor.ActorSystem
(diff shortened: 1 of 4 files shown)


smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/client/ServiceAccessLayer.scala
```diff
@@ -43,10 +43,17 @@ class ServiceAccessLayer(val baseUrl: URL)(implicit actorSystem: ActorSystem) {
...
   def serviceStatusPipeline: HttpRequest => Future[ServiceStatus] = sendReceive ~> unmarshal[ServiceStatus]
+  def getDataSetByUuidPipeline: HttpRequest => Future[DataSetMetaDataSet] = sendReceive ~> unmarshal[DataSetMetaDataSet]
 
   val statusUrl = toUrl("/status")
 
   def getStatus: Future[ServiceStatus] = serviceStatusPipeline {
     Get(statusUrl)
   }
+
+  // FIXME this should take either an Int or a UUID, but how?
```
natechols: So how is this done?  I couldn't figure out how to make Either[Int, UUID] work

smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/tools/PbService.scala
```diff
@@ -0,0 +1,110 @@
...
+    } children(
+      arg[Int]("dataset-id") required() action { (i, c) =>
+        c.copy(dataset_id = i)
+      } text "Dataset ID"
+    ) text "Show dataset details"
+  }
+}
+
+
+object PbServiceApp extends App {
```
natechols: This is just awful, I'm not sure what the right way to do this is (of course most of the actual functionality is missing too)

smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/client/ServiceAccessLayer.scala
mpkocher: The standard java/scala style is to use camelcase. We should try to follow that model.

mpkocher: There's a way to do this. However, for now can you use explicit getDataSetByUUID(uuid: UUID)

String sub can use s"${a}/$b" format.

natechols:
Tripped over .gitignore changes, will try again.

PR #24, opened 2016-05-04
bug 31728: add and populate createdBy field for jobs

skinner:
have job endpoints record and respond with the user that created the job, for jobs created by logged-in users


skinner:
having CI running on pull requests is awesome

PR #27, opened 2016-05-05
add job polling to dataset import

natechols:
plagiarized from PAWS


mpkocher:
This is a problem with indeterminism in the tests.

PR #28, opened 2016-05-05
port remaining functionality from python get-status tool

natechols:


smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/tools/GetStatus.scala
```diff
@@ -59,19 +71,30 @@ object GetStatusRunner extends LazyLogging {
...
-
-    val results = Await.result(fx, 5 seconds)
-    val (status) = results
-    println(status)
+    var xc = 1
+    var ntries = 0
+    while (ntries < c.maxRetries) {
+      ntries += 1
+      Thread.sleep(c.sleepTime)
+      val result = Try { Await.result(sal.getStatus, 5 seconds) }
```
natechols: What actually happens when the connection fails (e.g. typo in hostname) is that the "retries" get executed concurrently.  How do I force them to be executed serially instead?

smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/tools/GetStatus.scala
```diff
@@ -59,19 +71,30 @@ object GetStatusRunner extends LazyLogging {
...
+      result match {
+        case Success(x) => {
+          println(s"GET ${url}: SUCCESS")
+          println(x)
+          n = c.maxRetries
```
skinner: `val n`?

natechols: Huh.  Did I not compile this first?  One moment...

smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/tools/GetStatus.scala
skinner: AFAIK, Await *is* the way to make them execute serially; I'm not sure what's happening here.

skinner: Ah, OK, they *are* executing sequentially--it just doesn't look like it because the sleep time is shorter than you're expecting.  And the sleep time is short because Thread.sleep takes an argument in milliseconds.  Add the * 1000 and I'll merge the PR.

skinner: Also, can we make the sleep happen only between a failure and a retry?

skinner:
I talked with Michael about this briefly--he thought that you should be using recursion rather than the while loop.  IMO what you have works, so I think I'll go ahead and merge the PR.  But the way ntries works (like setting it equal to maxRetries in order to break out of the loop) seems a little clunky.  If you're open to the idea, try a recursive version and see if it seems like a simplification.

The failure in circleci doesn't look related to this change, so I'm not going to block this PR on that.

natechols:
It's definitely clunky - I found some better examples here:
http://stackoverflow.com/a/7931459

skinner:
Ah, those are nice.  Also, minor style point that I forgot about when I commented earlier, since you have `import scala.concurrent.duration._` you could say `c.sleepTime.seconds.toMillis` instead of `c.sleepTime * 1000`.

PR #30, opened 2016-05-07
more service endpoints

natechols:


smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/tools/GetStatus.scala
```diff
@@ -93,6 +93,20 @@ object GetStatusRunner extends LazyLogging {
         }
       }
     }
+    if (xc == 0) {
+      // FIXME this crashes with servers that have many jobs
```
natechols: not sure what to do about this...

mpkocher:
The test failures seem to be unrelated to these changes.

PR #34, opened 2016-05-12
add get-job command, and handle Int or UUID flexibly for jobs and datasets

natechols:


mpkocher:
Here's an example of using a more general Id format.

https://gist.github.com/mpkocher/8d37114680b9c39d9adb119f74ce76e1

PR #39, opened 2016-05-13
bug 32933: consolidate and increase timeouts

skinner:


skinner:
The test failure looks like the same intermittent CleanupSpec failure that I'm pretty sure is unrelated.

PR #42, opened 2016-05-13
Logging improvements

### LINKED ISSUE #18: Better Commandline tool logging model
Programmatically define and setup the logger for commandline tools. 

Currently it's using the application.conf, which is not the best model.

jfalkner:
Fixes #18 

> WIP: Will remove when done. Wanted to get the first pass of the code up before end of this week.

New logging related flags have been added to the CLI tools. This includes removing the `debug` flag and encapsulating it in the `LoggingConfig` mixin.

```bash
$ ./smrt-server-tools/target/pack/bin/get-smrt-server-status -h
Get SMRTLink status  0.1.0
Usage: get-status [options]

Tool to check the status of a currently running smrtlink server
...
  --debug
  --loglevel <value>
  --logfile <value>
  --logback <value>
```

Stuff left TODO:

- [ ] Cleanup the [Readme.md](https://github.com/PacificBiosciences/smrtflow/tree/logging_improvements/smrt-server-logging) so it matches the latest and explains this well.
- [x] Consider disabling Akka logging by default. It loads up a custom StandardOutputLogger based on `application.conf`
- [ ] Tweak the other CLI App classes to use the new LogConfig and LazyLogger
- [x] Make sure @skinner and @natechols confirm the build can use the new `--logfile` and `--loglevel` flags instead of the properties file.

Verbosity is much less now. Here is an example. Note the Akka logger makes those top 3 messages and controlling it appears to only be done via application.conf.

### Default, No Logging Args
By default the logging code will configure itself via LoggerConfigurator and silence most messages.

```bash
$ ./smrt-server-tools/target/pack/bin/get-smrt-server-status --host smrtlink-bihourly --port 8081 
```
```
[DEBUG] [05/16/2016 14:05:25.662] [main] [EventStream] StandardOutLogger started
[DEBUG] [05/16/2016 14:05:25.821] [main] [EventStream(akka://get-status)] logger log1-Slf4jLogger started
[DEBUG] [05/16/2016 14:05:25.822] [main] [EventStream(akka://get-status)] Default Loggers started
URL: http://smrtlink-bihourly:8081
GET http://smrtlink-bihourly:8081: SUCCESS
ServiceStatus(pacbio.smrtservices.smrtlink_analysis,Services have been up for 1 hour, 1 minute and 34.794 seconds.,3694794,6f5d51b3-ad66-4a40-9062-fca2f43e67e9,0.50.21-SNAPSHOT,secondarytest)
found endpoint /secondary-analysis/job-manager/jobs/import-dataset
...
Exiting pbscala.tools.get_status v0.1.0 with exit code 0
```

### Debug Mode (--debug)
This mode sends output to System.out. Convenient for seeing all logs or piping output somewhere. Can use `--loglevel` to change verbosity.

```bash
$ ./smrt-server-tools/target/pack/bin/get-smrt-server-status --host smrtlink-bihourly --port 8081 --debug
```

Sends `ERROR` level messages to the console by default. There are none.
```
[DEBUG] [05/16/2016 13:32:15.024] [main] [EventStream] StandardOutLogger started
2016-05-16  INFO [get-status-akka.actor.default-dispatcher-3] a.e.s.Slf4jLogger - Slf4jLogger started
[DEBUG] [05/16/2016 13:32:15.183] [main] [EventStream(akka://get-status)] logger log1-Slf4jLogger started
[DEBUG] [05/16/2016 13:32:15.185] [main] [EventStream(akka://get-status)] Default Loggers started
2016-05-16 17:32:15.183UTC DEBUG[get-status-akka.actor.default-dispatcher-3] a.e.EventStream - logger log1-Slf4jLogger started
URL: http://smrtlink-bihourly:8081
...
```

Changing the `--loglevel` to `DEBUG` shows more verbosity.

```bash
$ ./smrt-server-tools/target/pack/bin/get-smrt-server-status --host smrtlink-bihourly --port 8081 --debug --loglevel DEBUG
```

### File Mode (--logFile)
This mode sends output to the given log. Convenient for saving all logs when running the service related code or any other case where output should be saved but not sent to `System.out`.

```bash
# Append log messages at default level (ERROR) to /tmp/example.log
$ ./smrt-server-tools/target/pack/bin/get-smrt-server-status --host smrtlink-bihourly --port 8081 --logFile /tmp/example.log
```

```bash
# Log only WARN and worse messages to /tmp/example.log
$ ./smrt-server-tools/target/pack/bin/get-smrt-server-status --host smrtlink-bihourly --port 8081 --logfile /tmp/example.log --loglevel WARN
```

### Logback Mode (--logbackFile)
The code relies on SLFJ4 and Logback and originally a bunch of `logback.xml` files. If a use case requires more comprehensive config via `logback.xml` the `--logbackFile` param may be used. The value is the name of the XML file to parse.

```bash
$ ./smrt-server-tools/target/pack/bin/get-smrt-server-status --host smrtlink-bihourly --port 8081 --logbackFile logback.xml
```


smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/tools/GetStatus.scala
```diff
@@ -119,7 +119,7 @@ object GetStatusRunner extends LazyLogging {
 
 object GetStatusApp extends App with GetStatusParser {
   def run(args: Seq[String]) = {
-    val exitCode = parser.parse(args, DEFAULT) match {
+    val exitCode = parser.parse(LogConfig.trim(args), DEFAULT) match {
```
jfalkner: This is essentially how the new code allows command-line args that support the common use cases I could identify. Any `App` can use it. We have a fairly rigid arg parser and conversion from `Seq[String]` to a Config class, meaning the logging config can't easily live in the `apply` methods associated with `App` classes. If it did, it'd require updating many Config classes to reflect optional debug args.

skinner:
Nice job documenting!  Even incomplete, it's good to see.

re: akka logging, you can configure akka to log to slf4j; see the [application.conf](https://github.com/PacificBiosciences/smrtflow/blob/025c44b58c7c3f8133f10138f758005d3ef8a2e8/smrt-server-analysis/src/main/resources/application.conf#l12) in smrt-server-analysis/src/main/resources.

If you're going to delete the service logback config files, then I think you should also move the services into your new world.  We're using LazyLogging in a lot of places.  It looks like a lot of the files that use LazyLogging will only need an import change (which is nice!), but there are a fair number of them:
```
❯ grep -r LazyLogging */src/main | grep import | wc -l
      59
```

Moving from the properties file to the command line switches ought to be straightforward, I think; we'd have to change [start](http://swarm/files/depot/software/smrtanalysis/services-ui/pbservice-bundler/smrtlink_services_ui/bin/start) and [apply-config](http://swarm/files/depot/software/smrtanalysis/services-ui/pbservice-bundler/smrtlink_services_ui/bin/apply-config)

We just have to coordinate the start and apply-config changes with the github changeover.

smrt-server-logging/src/main/scala/com/pacbio/logging/Logger.scala
```diff
@@ -0,0 +1,180 @@
...
+import org.slf4j.LoggerFactory
+
+import scala.collection.JavaConversions._
+import scala.collection.mutable.ListBuffer
+
+/**
+  * Defines `logger` as a lazy value initialized with an underlying `org.slf4j.Logger`
+  * named according to the class into which this trait is mixed.
+  */
+trait LazyLogging {
```
mpkocher: What is the adding that is different that LazyLogging that the scala package is providing?

https://github.com/typesafehub/scala-logging/blob/master/src/main/scala/com/typesafe/scalalogging/Logging.scala#L27

smrt-server-logging/src/main/scala/com/pacbio/logging/Logger.scala
```diff
@@ -0,0 +1,180 @@
...
+  * See Readme.md for details about the shared parameters and examples of use.
+  */
+object LogConfig {
+
+  val consolePattern = "%X{akkaTimestamp} %-5level[%thread] %logger{0} - %msg%n"
+  val filePattern = "%date{yyyy-MM-dd} %X{akkaTimestamp} %-5level[%thread] %logger{1} - %msg%n"
+  val fileRollingPosfix = ".%i.gz"
+
+  // all the expected loggers
+  val loggerNames = "akka" :: "Slf4jLogger" :: "scala.slick" :: "spray" :: "log1-Slf4jLogge" :: Nil
```
mpkocher: Is there a style reason for not using Set("akka", "spray", "scala.slick")

And is "log1-Slf4jLogge" a typo?

smrt-server-logging/src/main/scala/com/pacbio/logging/Logger.scala
jfalkner: None other than removing a dependency. I cleaned up the code and removed the duplicate `LazyLogger`.

Originally, I assumed customizing this `trait` would be helpful. Incorrect assumption. No need to change it. Default SLFJ4 is used. FWIW, as far as I can tell we don't use any of the fancy features of scalalogging, albeit no great reasons to clean up the dependency.

smrt-server-logging/src/main/scala/com/pacbio/logging/Logger.scala
jfalkner: Yes, typo. Thanks for noting. Fixed in c984dd0d8214a90bb79df1b8e81db98774248dc7

No good reason for using `::` vs set. Order doesn't matter. Swapped to use `Set` as implied in c984dd0d8214a90bb79df1b8e81db98774248dc7

jfalkner:
>If you're going to delete the service logback config files, then I think you should also move the services into your new world.  We're using LazyLogging in a lot of places.  It looks like a lot of the files that use LazyLogging will only need an import change (which is nice!), but there are a fair number of them:
```
❯ grep -r LazyLogging */src/main | grep import | wc -l
      59
```

I failed to realize where all scalalogging was being used. I've scaled back my changes to focus on the services that we expect to leave log files. The code in `smrt-server-logging` can be reused elsewhere but I think wise to roll it out in a measured way. Start with the offending use case.

jfalkner:
> - [ ] Consider disabling Akka logging by default. It loads up a custom StandardOutputLogger based on `application.conf`

Skipping this for now. Akka adds three stdout messages when it loads and they appear to be outside normal logging mechanisms. I tweaked the config some but couldn't easily sort out the cause. Can look closer at it in later PRs. As-is, this PR removes the bulk of the unhelpful, verbose log messages. It also adds a common way to configure debugging via CLI apps.

mpkocher:
After the 3.1.x timeframe, I think we should rename pb-common-models to pb-common, then fold this back in. I would like to keep the root of the project to have a minimal number of subproject (if possible).

Do this work for you?

jfalkner:
> I think we should rename pb-common-models to pb-common, then fold this back in. I would like to keep the root of the project to have a minimal number of subproject (if possible).

I'm flexible. Post 3.1.x timeframe makes sense for further refactoring.

My preference is to favor smaller sub-projects, keep them in the repo for a few months and then break them out of the bigger repo as much as possible. Load as a versioned dependency JAR -- presumably a slow changing API. I do realize the idea with the SBT multi-project is to try and consolidate the build some and that isn't a great fit for that theme. My preference comes from wanting code that can't easily be deleted in favor of improvements. I find erring on the side of more repos greatly simplifies this task. Even it seems we're making a repo that has just a handful of Scala files.

Flexible though. Not a stickler about the above.

jfalkner:
I've started a branch to make it obvious what `App` files end up with the new logging code. Easy PR branch from this to review. Will post shortly.

skinner:
I think maybe I was making a mistaken assumption here:

> If you're going to delete the service logback config files, then I think you should also move the services into your new world. We're using LazyLogging in a lot of places. It looks like a lot of the files that use LazyLogging will only need an import change (which is nice!),

I assumed that importing and using your LazyLogging was how code that wanted to log would use your configuration.  But given Michael's question about it, and since you've deleted it, I guess that's not the case.  I'm still hazy on exactly how a piece of code that wants to write to the log ends up using your configuration here, can you connect those dots for me?

My underlying concern was that this PR deletes the logback config files, and it seems like that leaves smrt-server-base, smrt-server-link, and smrt-server-analysis without a logging configuration.  Is that right?  If so, I think it's a problem--what's the plan there?

jfalkner:
@skinner I took a stab at moving other CLI code to use the new logging in #42 with a hopefully helpful example comment [here](https://github.com/PacificBiosciences/smrtflow/pull/44/files#r63416086).

Above is how the source-code is updated to use the new conventions. At runtime the code now relies on the flags noted in the [smrt-server-logging/Readme.md](https://github.com/PacificBiosciences/smrtflow/blob/logging_improvements/smrt-server-logging/Readme.md). In short, these two ways (assuming you think it'll work).

1. Anything that should log to a file now has `--logfile` specified with the location to use and `--loglevel` with the verbosity. Similar location and same level as the old [logback.properties](https://github.com/skinner) file you noted.
2. If special logback.xml files are needed, they can override all the conventions and be passed to a command-line program via the `--logback` parameter.

Thoughts?

PR #43, opened 2016-05-14
more useful and human-readable feedback

natechols:


smrt-server-tools/src/main/scala/com/pacbio/secondary/smrttools/client/ServiceAccessLayer.scala
```diff
@@ -254,19 +254,24 @@ class ServiceAccessLayer(val baseUrl: URL)(implicit actorSystem: ActorSystem) {
     Get(toJobResourceUrl(JobTypes.PB_PIPE, jobId, ServiceResourceTypes.ENTRY_POINTS))
   }
 
-  private def getJobDataStore(jobType: String, jobId: Int) : Future[PacBioDataStore] = getDataStorePipeline {
+  // FIXME I think this is unmarshalling incorrectly
```
natechols: The problem here is currently: "Object is missing required member 'uniqueId'".  But using PacBioDataStore didn't work either - it complaned about "createdAt" (and that seems inappropriate anyway).

mpkocher: So the call is something like this?

http://smrtlink-beta:8081/secondary-analysis/job-manager/jobs/pbsmrtpipe/6/datastore

mpkocher: I believe it needs to be DataStoreServiceFile.

By convention, there's often a non-service data model, then a wrapped service data model. For example the job options and the datastore file follow this pattern.

natechols: That worked.

PR #44, opened 2016-05-16
Services with updated logging

jfalkner:
Simple PR branched from #42 so that it is easy to see what services now have the improved logging.

- `GetStatusService` see #42 -- was changed in that PR
- `PbService`
- `DataSetMergerTool`
- `FastaToReference`
- `FastaToReferenceDataSet`
- `ImportDataSet`
- `MovieMetaDataToDataSet` 
- `MovieMetaDataToDataSetRtc` 
- `ReferenceInfoToDataSet` 
- `ValidateDataSet`

Looks like 11 candidates.

```
grep -r --include=*.scala "new OptionParser" . | wc -l
11

$ grep -r --include=*.scala "new OptionParser" .
./smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/tools/DataSetMergerTool.scala:  val parser = new OptionParser[DataSetMergerOptions]("validate-datasets") {
./smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/tools/FastaToReference.scala:  val parser = new OptionParser[FastaToReferenceConfig]("fasta-to-reference") {
./smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/tools/FastaToReferenceDataSet.scala:  val parser = new OptionParser[FastaToReferenceDataSetConfig]("fasta-to-dataset") {
./smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/tools/ImportDataSet.scala:  val parser = new OptionParser[ImportDataSetConfig]("import-dataset") {
./smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/tools/MovieMetaDataToDataSet.scala:  val parser = new OptionParser[MovieMetaDataToDataSetConfig]("movie-metadata-to-dataset") {
./smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/tools/MovieMetaDataToDataSetRtc.scala:  val parser = new OptionParser[MovieMetaDataToDataSetRtcConfig]("movie-metadata-to-dataset-rtc") {
./smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/tools/PbDataSetTools.scala:  lazy val parser = new OptionParser[CustomConfig]("pb-datasets") {
./smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/tools/ReferenceInfoToDataSet.scala:  val parser = new OptionParser[ReferenceConverterConfig]("reference-to-dataset") {
[...]

PR #45, opened 2016-05-16
Logging improvements (Again)

jfalkner:
CC @skinner @mpkocher 

A few commits didn't make it in before #42 was merged. I also would like to merge #44 to the branch before `master` so that the tool refactoring goes in.

I'll keep this up for some feedback then merge.


jfalkner:
Merging this since it is just a few doc fixes and #42 was merged.

Issue #50, opened 2016-05-17
Tests now dump too much logging info


jfalkner:
CC @skinner @natechols 

@mskinner noted that running `sbt test` now makes over 50,000 lines of output. It shouldn't and didn't previously. This is after #42 was merged.

The suspected cause is that the code isn't being configured by the old `logback.xml` file that was deleted.  Adding to this, the new code does do a default config via `Configurator`; however, that relies on having the `smrt-server-logging` JAR in classpath. The fix here may be as simple as adding this JAR as needed.

How to replicated the issue: `sbt test`

A quicker test being used to debug this issue.

```bash
# Run one test that makes INFO level log messages. Too verbose.
sbt testOnly SanitySpec &> SanitySpec.output
# Count the total lines.
cat SanitySpec.output | wc -l
# Count lines we know shouldn't be included.
grep "INFO" SanitySpec.output | wc -l

...
$ cat SanitySpec.output | wc -l
     450
$ grep "INFO" SanitySpec.output | wc -l
       4
```

jfalkner:
@skinner noted in [#pbservices](https://pacbio.slack.com/messages/pbservices/files/) that adding the JAR seems to work. I'll wrap this issue for completeness and add that change.

> given your explanation (which helps, thanks; that’s what I was asking for in one of my comments on PR 42) I tried adding logging to the dependsOn for smrtServerBase, smrtServerLink, and smrtServerAnalysis, and that fixed it

PR #51, opened 2016-05-17
Logging JAR now in all projects

### LINKED ISSUE #49: Verbose Slick logging
When running the tests, or when running `sbt smrt-server-analysis/assembly`, then `java -jar` I am seeing  a lot of slick logging. 

The logging is currently far too verbose, we would need to address this before deploying.

jfalkner:
Fixes #49 and #50 by adding the `smrt-server-logging` JAR to classpath, which means the default `Configurator` will be used.

Much less verbose test output.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -124,31 +124,31 @@ lazy val common = (
 // "pbscala" or pacbio-secondary in perforce repo
 lazy val smrtAnalysis = (
   PacBioProject("smrt-analysis")
-    dependsOn (common)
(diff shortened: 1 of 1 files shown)


jfalkner:
Tested by hand and confirmed that the log verbosity changed as expected. Matched @skinner's observation noted in Slack.

jfalkner:
Note, I had to `sbt clean; sbt packInstall` before running tests to get the new JAR

PR #59, opened 2016-05-18
SecondaryAnalysisApp CLI logging arg support

### LINKED ISSUE #55: enable smrt-server-logging configuration in the server
We used to supply a properties file at server startup time, and the logback.xml files would reference those properties.  Now that the logback.xml files are gone, we don't have a way for the services to take logging configuration like the file to log to.

Jayson has added command-line parameter processing for the command-line tools.  But for the server programs, I'm not sure if the command line is the nice way to do it.  Maybe com.pacbio.logging.LoggerConfigurator could look at system properties and use them?  Or maybe it could load our existing properties file if we pass that as a system property?

jfalkner:
@skinner @mpkocher 

Fixes #55

Adds command-line argument parsing for logging to the servers noted here https://github.com/PacificBiosciences/smrtflow/issues/55#issuecomment-219824672

The server code didn't use `scopt`; however, it is an `App` entry point that has CLI args. I added a util method so that if logging options are passed to the command line, they'll be used. This means that command line `--logfile` and `--loglevel` flags can now be used instead of a `logback.xml` file plus a properties file to config the log file and log level.

e.g.

```bash
java -jar target/scala-2.11/smrt-server-analysis*.jar --logfile /tmp/my_file.log --loglevel ERROR
```

And `-h` still works.

```
Usage: ./app_with_logging [options]

This is an app that supports PacBio logging flags. 
  -h | --help
        Show Options and exit
  --debug
  --loglevel <value>
  --logfile <value>
  --logback <value>
```

### Require Log File
These servers should have a log file specified during runtime. I added a utility method named `parseRequireFile` so that by default the code will not run without a log file. 

```
# Old way the code was run. Relied on config files that no longer exist.
java -jar target/scala-2.11/smrt-server-analysis*.jar

You must set the logger output with a --logfile parameter
  e.g.
      java -jar my_code.jar --logfile example_file.log
```

We can use `/dev/null` for dev work, if you don't want to log to a file.

### DIFF
--- a/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/appcomponents/SecondaryAnalysisApp.scala
+++ b/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/appcomponents/SecondaryAnalysisApp.scala
@@ -69,7 +69,7 @@ object SecondaryAnalysisServer extends App with BaseServer with SecondaryApi {
   override val host = providers.serverHost()
   override val port = providers.serverPort()
 
-  LoggerOptions.parse(args)
+  LoggerOptions.parseRequireFile(args)
 
   start
 }
--- a/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/appcomponents/SecondaryAnalysisApp.scala
+++ b/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/appcomponents/SecondaryAnalysisApp.scala
@@ -6,6 +6,7 @@ import com.pacbio.common.app.{BaseServer, BaseApi}
 import com.pacbio.common.dependency.{TypesafeSingletonReader, Singleton}
 import com.pacbio.common.logging.{LoggerFactoryProvider, LogResources}
 import com.pacbio.common.models.LogResourceRecord
+import com.pacbio.logging.LoggerOptions
 import com.pacbio.secondary.smrtlink.app.SmrtLinkProviders
 import com.pacbio.secondary.smrtlink.auth.SmrtLinkRolesInit
 import com.pacbio.secondary.smrtserver.services._
@@ -67,6 +68,8 @@ trait SecondaryApi extends BaseApi with SmrtLinkRolesInit with LazyLogging {
 object SecondaryAnalysisServer extends App with BaseServer with SecondaryApi {
   override val host = providers.serverHost()
   override val port = providers.serverPort()
+
+  LoggerOptions.parse(args)
 
   start
 }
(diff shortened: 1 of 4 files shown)


jfalkner:
@skinner After looking at the code, it appeared that adding logging command line option support to the servers was straight-forward. I went with that approach. My logic was as follows.

1. Nice to have consistent logger config use. Learn once, use everywhere.
2. Explicit passing of `--logfile` and `--loglevel` makes it obvious as to where logs will appear in the same spot that the code is launched. Seems no need for the magic(ish) config of a `logback.xml` plus related properties file in the classpath.

jfalkner:
Looks like the bug I was seeing was just in the IDE. When running your exact commands, it seems to launch correctly and with a warning.

```
Warning. Assuming relative path /Users/jfalkner/tokeep/git/PacificBiosciences/smrtflow/smrt-server-analysis/jobs-root
[WARNING] Unable to load CMD template from pb-engine.pb-cmd-template Error No configuration setting found for key 'pb-engine.pb-cmd-template'
```

jfalkner:
Tested smrt-server-analysis via command line.

```bash 
# Confirm no-arg or lacking --logfile fails.
$ java -jar target/scala-2.11/smrt-server-analysis*.jar
You must set the logger output with a --logfile parameter
  e.g.
      java -jar my_code.jar --logfile example_file.log


# -h works even sans --logfile
$ java -jar target/scala-2.11/smrt-server-analysis*.jar -h
Warning: Unknown argument 'target/scala-2.11/smrt-server-analysis_2.11-0.1.1-SNAPSHOT.jar'
Usage: ./app_with_logging [options]

This is an app that supports PacBio logging flags. 
  -h | --help
        Show Options and exit
  --debug
  --loglevel <value>
  --logfile <value>
  --logback <value>


# Expected use works when the `--logfile` param is passed
$ java -jar target/scala-2.11/smrt-server-analysis*.jar --logfile todelete_log.log
...
```

jfalkner:
@skinner should be ready for merge. I'll hold off in hopes that you can take a look today. If LGTY, please merge away and we can start using it.

mpkocher:
@natechols or @skinner  This will require changes to pbbundler, specifically apply-conf to set write the logback.xml file with the required path.

mpkocher:
Looks good. @skinner does this work for you?

skinner:
I'd like for the no-arg server behavior to be whatever default logging we decide on for server development.

The deployed server is usually started by script, so I don't think the command-line ergonomics matter much in the deployed case.  But I (and I assume other devs) run the server from the command line regularly.  So I'd like us to optimize the server command-line defaults for dev usage.  IIRC the old behavior of the server programs launched without command-line params was to log to stdout at the INFO level; that seems like a reasonable no-arg default to me.

PR #60, opened 2016-05-18
Lazy make needed  `db` dir

### LINKED ISSUE #58: Cleanup db/* references in smrt-server-analysis
CC @mpkocher @skinner 

Split out from this comment https://github.com/PacificBiosciences/smrtflow/issues/55#issuecomment-219853372. It seems we have some hard-coded path logic related to using an sqlite db.

Run the code with the below or directly from IntelliJ.

```bash
sbt smrt-server-analysis/assembly
cd smrt-server-analysis
mkdir db
java -jar target/scala-2.11/smrt-server-analysis*.jar
```

Output will error with similar to the following.

```bash
Connected to the target VM, address: '127.0.0.1:57556', transport: 'socket'
Warning. Assuming relative path /Users/jfalkner/tokeep/git/PacificBiosciences/smrtflow/jobs-root
[WARNING] Unable to load CMD template from pb-engine.pb-cmd-template Error No configuration setting found for key 'pb-engine.pb-cmd-template'
org.flywaydb.core.api.FlywayException: Unable to obtain Jdbc connection from DataSource (jdbc:sqlite:db/analysis_services.db) for user '': path to 'db/analysis_services.db': '/Users/jfalkner/tokeep/git/PacificBiosciences/smrtflow/db' does not exist
	at org.flywaydb.core.internal.util.jdbc.DriverDataSource.getConnectionFromDriver(DriverDataSource.java:407)
	at org.flywaydb.core.internal.util.jdbc.DriverDataSource.getConnection(DriverDataSource.java:365)
	at org.flywaydb.core.internal.util.jdbc.JdbcUtils.openConnection(JdbcUtils.java:50)
	at org.flywaydb.core.Flyway.execute(Flyway.java:1314)
	at org.flywaydb.core.Flyway.migrate(Flyway.java:917)
	at com.pacbio.secondary.smrtlink.actors.JobsDao.initializeDb(JobsDao.scala:935)
	at com.pacbio.secondary.smrtserver.appcomponents.SecondaryApi$class.startup(SecondaryAnalysisApp.scala:49)
	at com.pacbio.secondary.smrtserver.appcomponents.SecondaryAnalysisServer$.startup(SecondaryAnalysisApp.scala:67)
	at com.pacbio.common.app.BaseApi$ServiceActor.preStart(BaseSmrtServerApp.scala:140)
	at akka.actor.Actor$class.aroundPreStart(Actor.scala:470)
	at com.pacbio.common.services.package$RoutedHttpService.aroundPreStart(package.scala:169)
	at akka.actor.ActorCell.create(ActorCell.scala:580)
	at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:456)
	at akka.actor.ActorCell.systemInvoke(ActorCell.scala:478)
	at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:263)
	at akka.dispatch.Mailbox.run(Mailbox.scala:219)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(AbstractDispatcher.scala:393)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
Caused by: java.sql.SQLException: path to 'db/analysis_services.db': '/Users/jfalkner/tokeep/git/PacificBiosciences/smrtflow/db' does not exist
	at org.sqlite.core.CoreConnection.open(CoreConnection.java:117)
	at org.sqlite.core.CoreConnection.<init>(CoreConnection.java:66)
	at org.sqlite.jdbc3.JDBC3Connection.<init>(JDBC3Connection.java:21)
	at org.sqlite.jdbc4.JDBC4Connection.<init>(JDBC4Connection.java:23)
	at org.sqlite.SQLiteConnection.<init>(SQLiteConnection.java:45)
	at org.sqlite.JDBC.createConnection(JDBC.java:114)
	at org.sqlite.JDBC.connect(JDBC.java:88)
	at org.flywaydb.core.internal.util.jdbc.DriverDataSource.getConnectionFromDriver(DriverDataSource.java:404)
	... 20 more
```

jfalkner:
Fixes #58 

I'm not sure if this is the ideal long-term fix, but it makes the code work as expected when running the code as noted in #58. The problematic case was if you lacked a `db` directory. Without manually making it, the code would error and exit.

### DIFF
--- a/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
+++ b/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
@@ -36,8 +36,11 @@ class Dal(val dbURI: String) {
   val flyway = new Flyway() {
     override def migrate(): Int = {
       // lazy make file database dir as needed
-      val file = scala.reflect.io.File(dbURI.stripPrefix("jdbc:sqlite:"))
-      if (file.parent != null && !file.parent.exists) file.parent.createDirectory()
+      val file = Paths.get(dbURI.stripPrefix("jdbc:sqlite:"))
+      if (file.getParent != null) {
+        val dir = file.getParent.toFile
+        if (!dir.exists()) dir.mkdirs()
+      }
 
       return super.migrate()
     }
--- a/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
+++ b/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
@@ -28,8 +28,9 @@ import scala.language.postfixOps
 import scala.slick.driver.SQLiteDriver.simple._
 import scala.slick.jdbc.meta.MTable
 import scala.util.{Failure, Success, Try}
-
 import org.flywaydb.core.Flyway
+
+import scala.reflect.io.Directory
 
 
 // TODO(smcclellan): Move this class into the c.p.s.s.database package? Or eliminate it?
@@ -932,6 +933,10 @@ with DataSetStore {
   var _runnableJobs = mutable.Map[UUID, RunnableJobWithId]()
 
   def initializeDb(): Unit = {
+    // lazy make ./db as needed
+    val dir = Directory("db")
+    if (!dir.exists) dir.createDirectory()
+
     dal.flyway.migrate()
   }
 }
--- a/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
+++ b/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
@@ -35,8 +35,8 @@ import org.flywaydb.core.Flyway
 class Dal(val dbURI: String) {
   val flyway = new Flyway() {
     override def migrate(): Int = {
-      // lazy make dev `db` database dir as needed. intended only for dev work
-      if (dbURI == "jdbc:sqlite:db/analysis_services.db") {
+      // lazy make directories as needed for sqlite
+      if (dbURI.startsWith("jdbc:sqlite:")) {
         val file = Paths.get(dbURI.stripPrefix("jdbc:sqlite:"))
         if (file.getParent != null) {
           val dir = file.getParent.toFile
--- a/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
+++ b/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
@@ -30,12 +30,18 @@ import scala.slick.jdbc.meta.MTable
 import scala.util.{Failure, Success, Try}
 import org.flywaydb.core.Flyway
 
-import scala.reflect.io.Directory
-
 
 // TODO(smcclellan): Move this class into the c.p.s.s.database package? Or eliminate it?
 class Dal(val dbURI: String) {
-  val flyway = new Flyway()
+  val flyway = new Flyway() {
+    override def migrate(): Int = {
+      // lazy make file database dir as needed
+      val file = scala.reflect.io.File(dbURI.stripPrefix("jdbc:sqlite:"))
+      if (file.parent != null && !file.parent.exists) file.parent.createDirectory()
+
+      return super.migrate()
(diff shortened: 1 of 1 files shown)


jfalkner:
@mpkocher @skinner  any downside to this lazy-make approach? From my initial read of the code, it seems OK, but I don't know the intended use of the sqlitedb. 

If LGTY, ready to merge.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
```diff
@@ -28,9 +28,10 @@ import scala.language.postfixOps
 import scala.slick.driver.SQLiteDriver.simple._
 import scala.slick.jdbc.meta.MTable
 import scala.util.{Failure, Success, Try}
-
 import org.flywaydb.core.Flyway
 
+import scala.reflect.io.Directory
```
mpkocher: I think using java.nio.file would be the preferred model. There's also apache-commons-io to address use cases that aren't addressed by the java stdlib

jfalkner: I just updated the code to use a smarter lazy-make based on the param, not hard-coded `db`. I used scala.reflect.io.File though.

Should be straight-forward to switch to java.nio.file. Why the preference?

mpkocher:
http://stackoverflow.com/questions/20149397/is-it-better-to-use-import-scala-reflect-io-file-or-java-io-file-in-scala?answertab=active#tab-top

skinner:
AFAIK the code is pretty database-backend agnostic.  But this change makes some assumptions about sqlite and about having a jdbc uri with a path in it, which goes in the wrong direction in terms of backend-agnostic-ness.

I don't think the `db` thing matters that much.  Can we just add a db dir to git under smrt-server-analysis and smrt-server-link and move on?  That certainly works for me on the command line.  Would it work for your IDE?  What working directory does your IDE start the server process in?  If it doesn't start the server in the smrt-server-<x> directory that you're working on, we could just remove `db/` from the default jdbc uri and have the sqlite file one directory level higher.

Either of those solutions would avoid making those sqlite/file assumptions.

jfalkner:
Thanks @mpkocher. The SO doesn't give too many specifics but does have a line noting the caveats of the scala-reflect.jar. In general, it seems don't use that unless you've read up on it and know better.

I don't know better and java.nio.file works well. Updated to use it as suggested in https://github.com/PacificBiosciences/smrtflow/pull/60/commits/c5db2e387b2bfb369b0afd9d19afc3560aa2bdc4

skinner:
My bad, I remembered that perforce didn't store empty directories but I mistakenly thought that git did.  So never mind the empty-directory solution.  Maybe we can make it non-empty, though?

jfalkner:
> AFAIK the code is pretty database-backend agnostic. But this change makes some assumptions about sqlite and about having a jdbc uri with a path in it, which goes in the wrong direction in terms of backend-agnostic-ness.

Good point. In practice, I find it rare that code really is db agnostic or that you change dbs. For example, we'll probably use sqlite this was for awhile until it needs to be yanked in favor of something better.

> I don't think the db thing matters that much. Can we just add a db dir to git under smrt-server-analysis and smrt-server-link and move on?

Agreed on this one not being a big deal. I've tweaked the code to have a check that the dev db URI is being used. If not, the helper code will be skipped. I favor that over the db in git mainly because the git tweak assumes you'll alway run the dev server in the git repo location.

I'm good going either way though. If no one feels strongly, I'll go with the latest code that has the dev db check.

jfalkner:
Followup from a off-thread discussion with @skinner 

> proposal: test for the jdbc:sqlite: prefix and test for the directory existing, remove the “dev” test that does an equality test against the full URI, lazily create the dir if sqlite prefix && directory doesn’t exist

SGTM. Updated the code to match in https://github.com/PacificBiosciences/smrtflow/pull/60/commits/d49d58a1380e368b9a0f1ab406fed724ce755e61

PR #93, opened 2016-05-25
more client refactoring

natechols:


smrt-server-base/src/main/scala/com/pacbio/common/tools/PbService.scala
```diff
@@ -0,0 +1,114 @@
...
+import scala.io.Source
+
+import java.net.URL
+
+import com.pacbio.common.client._
+import com.pacbio.logging.{LoggerConfig, LoggerOptions}
+
+
+trait PbServiceBase {
+  object Modes {
```
natechols: This is still a problem - I need to be able to extend the available Modes in other clients.

Issue #89, opened 2016-05-25
Create smrt-server-lims and LimsSubreadDataSet


jfalkner:
A step in @mpkocher's design for rolling out a service that provides better LIMS data tracking and resolution of short IDs to proper ID. As a start, this will 

- Create a sub-project named `smrt-server-lims` that extends `smrt-server-link`
- Can start with a MVP of a lims.yml import web service that has a dedicated DB for it and skipping use of jobs. Later add in support for shared DB and jobs.
- Have the UUID be a nullable FK. It is fine that this data will be a super set of DataSubreadSet.

Followup work. Recast or otherwise post to a new issue after this finishes.

- com.pacbio.secondaryinternal.services.LimsResolverService to be in `smrt-server-lims`
- Deploy this is a stable spot so the service is available 24/7
- Add in support for resolving short, user-friendly identifiers to the correct ID for the 8 basic data types (classes with trait `ServiceDataSetMetadata`). Base this on use cases from Milhouse style runs. e.g LIMS ID -> subreadset UUID, reference -> subreadset UUID, ...

jfalkner:
Here is a link to the [`import-pbi-collections.py` in the gist with the ES code](https://gist.github.com/mpkocher/cc05927ab855d8275b34402877eaf0a1#file-import-pbi-collections-py). Examples of the path to find data and fields imported.

jfalkner:
Example `lims.yml` can be found in `/pbi/collections`. From OS X while on VPN you can use the following.

```bash
$ find /net/pbi/collections/322/3220001 -name "lims.yml"
/net/pbi/collections/322/3220001/r54003_20160212_165105/1_A01/lims.yml
/net/pbi/collections/322/3220001/r54003_20160212_172622/1_A01/lims.yml
/net/pbi/collections/322/3220001/r54003_20160212_181719/1_A01/lims.yml
...
```

jfalkner:
It turns out that the above examples aren't great. In chatting with @mpkocher, they may be from the RS era and things like the `uid` field don't make sense. The `path` is also not what is expected and should explicitly reference the subreadset, `/pbi/collections/322/3220001/r54003_20160212_165105/1_A01/m54003_160212_165114.subreadset.xml`.

I'm at a minor block to clarify some good input data and expected queries and output post-resolution. @mpkocher or anyone else, advice?

jfalkner:
Right now the RESTful endpoints are pretty simple in #119 

[...]

PR #96, opened 2016-05-26
`None` as the default PacBioComponentManifest dependency arg

jfalkner:
CC @mpkocher look good?

In working on #89 I realized `None` is passed many times when `PacBioComponentManifest` is created for services. Only 4 or so uses pass an arg. Making `None` tidies up the code a fair bit and avoids needing to put an (often unnamed) arg of `None`.

#### Old
```scala
case class PacBioComponentManifest(id: String, name: String, version: String, description: String, dependencies: Option[Seq[PacBioComponent]])
```
#### New
```scala
case class PacBioComponentManifest(id: String, name: String, version: String, description: String, dependencies: Option[Seq[PacBioComponent]] = None)
 ```

Which lets you do cleanup such as the following in over 20 spots in our codebase.

```scala
 val manifest = PacBioComponentManifest(toServiceId("smrtlink_resource_resolver"),
     "SMRT Link Resource Resolver", "0.1.0",
-    "Service to resolve job paths from job ids using SMRT Link Resources id.", None)
+    "Service to resolve job paths from job ids using SMRT Link Resources id.")
```

### DIFF
--- a/smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/services/LimsResolverService.scala
+++ b/smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/services/LimsResolverService.scala
@@ -2,7 +2,6 @@ package com.pacbio.secondaryinternal.services
 
 import com.pacbio.common.dependency.Singleton
 import com.pacbio.common.models.PacBioComponentManifest
-import com.pacbio.common.models.PacBioComponentManifest
 import com.pacbio.secondaryinternal.BaseInternalMicroService
 import com.pacbio.secondaryinternal.Constants
 import com.pacbio.secondaryinternal.InternalAnalysisJsonProcotols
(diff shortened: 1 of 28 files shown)


jfalkner:
Here is the `grep` I used to find all of these.

`grep -rh --include=*.scala -A 4 "val manifest =" . `

And similarly to confirm none are left.

` grep -rh --include=*.scala -A 4 "val manifest =" . | grep -h -A 4 None`

jfalkner:
Tests all pass.

smcclellan7:
Would it make even more sense to do this?

`dependencies: Seq[PacBioComponent] = Nil`

I'm not sure what the difference between None and Some(Nil) would be, anyway.

mpkocher:
I agree with Steve. If you want a default value, it should be Nil or Seq.empty[PacBioComponent].

jfalkner:
SGTM. I'll give it a try and double-check that the code isn't doing something that cares about `None`. It seems that `None` is intended to be a return value and `Nil` means empty set.

jfalkner:
I just switched to `dependencies: Seq[PacBioComponent] = Nil`. I'm not sure how everything originally ended up with `None`.

Tests still pass and it appears that we don't do much with dependencies right now. The only use I could find was to return a list of the manifests.

PR #101, opened 2016-05-27
Fix failure case which was causing the system not stop processing jobs

mpkocher:


mpkocher:
@natechols This might warrant updating smrtlink-beta with mainline.

PR #104, opened 2016-05-28
Mk add makefile

mpkocher:


smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/ImportDataSetJob.scala
```diff
@@ -72,7 +72,18 @@ with MockJobUtils with timeUtils {
 
       val logPath = job.path.resolve("pbscala-job.stdout")
 
-      val reportFiles = DataSetReports.runAll(srcP, dst, job.path, jobTypeId, resultsWriter)
+      // This should never stop a dataset from being imported
```
mpkocher: @skinner Can you comment on the current model of when dataset importing would fail if the job fails?

PR #107, opened 2016-05-31
Load version from existing version.properties

### LINKED ISSUE #84: Add git sha to Version string exposed in /status
properties.version has the git sha, we should append the version string in /status with this value

jfalkner:
Fixes #84 

Loading property files in scala seems to rely on the same old (slightly) cumbersome Java `Properties` and `load` method. I updated our `buildVersion` method to do that.

```bash
# remake the needed JAR
sbt clean smrt-server-analysis/assembly smrt-server-analysis/run

# check the version via URL
curl http://localhost:8070/status

{
  "uuid": "f18647c4-591a-462c-8867-96fc28c26505",
  "version": "af4a30e670a76f3a38e2fdc420455c830e0dc479",
  "id": "pacbio.smrtservices.smrtlink_analysis",
  "uptime": 4388,
  "message": "Services have been up for 4.388 seconds.",
  "user": "jfalkner"
}
```

I couldn't find a convenient Scala way to walk the ClassLoader tree, look for `properties` files with a given name and give the first one back as a map. I may cook this up for fun.

### DIFF
--- a/smrt-server-base/src/main/scala/com/pacbio/common/actors/StatusServiceActor.scala
+++ b/smrt-server-base/src/main/scala/com/pacbio/common/actors/StatusServiceActor.scala
@@ -1,11 +1,13 @@
 package com.pacbio.common.actors
 
-import java.util.UUID
+import java.util.{Properties, UUID}
 
-import akka.actor.{Props, ActorRef}
+import akka.actor.{ActorRef, Props}
 import com.pacbio.common.dependency.Singleton
 import com.pacbio.common.time.{Clock, ClockProvider}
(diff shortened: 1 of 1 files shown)


jfalkner:
Tests pass. I'll merge when CircleCI is green.

I realized that there may be one failure mode where `version.properties` exists but lacks a `version` key. Since we make the file, this shouldn't happen. If it does, I'll improve the code.

mpkocher:
So the explicit version in the services should be of the form {version}-{sort-sha}, e.g., "0.1.2-af4a30e"

jfalkner:
I just reopened #84 and I'll have a PR to match the new format shortly.

PR #111, opened 2016-06-01
Logger default level INFO and reduce Slick logging verbosity

### LINKED ISSUE #103: Remove Verbose slick logging from running from sbt

jfalkner:
CC @skinner @mpkocher 

Fixes #103

This PR updates the default logging behavior to match the [clarified expectations](https://github.com/PacificBiosciences/smrtflow/issues/103#issuecomment-222753670) here. The change also ends up reducing a chunk of the verbose Slick logging #103 since it hides DEBUG level messages. I also improved the code to persist logging config state since sometimes the `Configurator` timing appears to sometimes change. e.g. debugger vs command-line

[README.md docs](https://github.com/PacificBiosciences/smrtflow/blob/logger_debug_default/smrt-server-logging/README.md) are updated too.

Can test by running the analysis server.

```bash
# rebuild the JAR needed for command-line use
sbt clean compile smrt-server-analysis/assembly

# run the analysis server from command line
sbt smrt-server-analysis/run

...
[info] 2016-06-01  INFO [smrtlink-analysis-server-akka.actor.default-dispatcher-2] c.p.s.s.a.SecondaryApi$$anon$1 - Loaded pb-engine.max-workers -> 35
[info] Warning. Assuming relative path /Users/jfalkner/tokeep/git/PacificBiosciences/smrtflow/smrt-server-analysis/jobs-root
[info] 2016-06-01  INFO [smrtlink-analysis-server-akka.actor.default-dispatcher-2] c.p.s.s.a.SecondaryApi$$anon$1 - Loaded pb-engine.pb-tools-env -> 
[info] 2016-06-01 16:08:07.024UTC INFO [smrtlink-analysis-server-akka.actor.default-dispatcher-3] c.p.s.a.e.a.EngineDaoActor - Starting DAO actor
[info] 2016-06-01 16:08:07.028UTC INFO [smrtlink-analysis-server-akka.actor.default-dispatcher-3] c.p.s.a.e.a.EngineManagerActor - Starting manager actor Actor[akka://smrtlink-analysis-server/user/engine-manager#987122468] with EngineConfig(35,,/Users/jfalkner/tokeep/git/PacificBiosciences/smrtflow/smrt-server-analys
```

And you can toggle verbosity as follows.

```bash
# only show errors
sbt "smrt-server-analysis/run --loglevel ERROR"
...
```

Similarly in files

```bash
sbt "smrt-server-analysis/run --logfile my_file.log --loglevel ERROR"
```

### DIFF
--- a/smrt-server-logging/src/main/scala/com/pacbio/logging/LoggerOptions.scala
+++ b/smrt-server-logging/src/main/scala/com/pacbio/logging/LoggerOptions.scala
@@ -21,7 +21,7 @@ object LoggerOptions {
    */
   def add(parser: OptionParser[LoggerConfig]): Unit = {
 
-    parser.opt[Unit]("debug") action { (x, c) =>
+    parser.opt[Unit]("log2stdout") action { (x, c) =>
       c.configure(c.logbackFile, c.logFile, true, c.logLevel)
     } text "If true, log output will be displayed to the console. Default is false."
 
@@ -61,8 +61,8 @@ object LoggerOptions {
   }
 
   def parseAddDebug(args: Seq[String]): Unit = {
-    val requireOne = Set("--logfile", "--debug", "-h")
-    val v = if (args.filter(requireOne).isEmpty) args :+ "--debug" else args
+    val requireOne = Set("--logfile", "--log2stdout", "-h")
+    val v = if (args.filter(requireOne).isEmpty) args :+ "--log2stdout" else args
     parse(v)
   }
 }
--- a/smrt-server-logging/README.md
+++ b/smrt-server-logging/README.md
@@ -7,7 +7,7 @@ Simple logging for two common use cases in the PacBio tools. See [smrtflow#42](h
 ```bash
 $ ./smrt-server-tools/target/pack/bin/get-smrt-server-status -h
 ...
-  --debug
+  --log2stdout
         If true, log output will be displayed to the console. Default is false.
   --loglevel <value>
         Level for logging: "ERROR", "WARN", "DEBUG", or "INFO". Default is "ERROR"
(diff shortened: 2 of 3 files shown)


skinner:
Just to clarify, I think this was a typo (from [the bug comments](https://github.com/PacificBiosciences/smrtflow/issues/103#issuecomment-223011857)):
> I'll get a PR up to switch to DEBUG as default

Looking at the code, it looks like this switches to INFO as default (not "DEBUG as default"), which is what I wanted.  Thanks!  +1 for merging.

smrt-server-logging/src/main/scala/com/pacbio/logging/LoggerOptions.scala
```diff
@@ -21,7 +21,7 @@ object LoggerOptions {
    */
   def add(parser: OptionParser[LoggerConfig]): Unit = {
```
mpkocher: In a future iteration, I would suggest adding an optional arg to set the default logging values, such as the log level.

jfalkner:
> Looking at the code, it looks like this switches to INFO as default (not "DEBUG as default"), which is what I wanted. Thanks!

Yep! INFO is the new default.

smrt-server-logging/src/main/scala/com/pacbio/logging/LoggerOptions.scala
jfalkner: I don't follow the difference between this request and the existing `--loglevel` flag that lets you set logging values. It is part of the default set of command-line params the logging code supports. An example or some more detail would be helpful.

mpkocher: Setting the default value (e.g., ERROR). In other words, set the default log level so that it's not needed from the commandline. 

This is useful for a tool that shouldn't have any stdout (or be launched in "quiet" mode by default).

PR #95, opened 2016-05-26
Scalastyle Checker: Part 1

jfalkner:
> Bumping the below todos to the next PR. This PR has a good chunk of the docs noted and cleans up a large portion of our code base to match two of the style rules.
>
> - [ ] Add doc on how to wire it to Intellij
> - [ ] Then a single commit to fix everything using this style format
>
> I'm currently pausing to think about how to do #2 above. The cleanup will be tedious and, if heading that route, I'd like to try and set a pre-commit hook or similar to enforce ongoing style.

Fixes #62 

I wanted to put up the start of using [Scalastyle.org's style checker](http://www.scalastyle.org/) to help enforce Scala's style rules and twitters [Effective Scala](http://twitter.github.io/effectivescala/#Formatting) guidelines. 

```bash
# run the checks
sbt scalastyle
```

IDE use in IntelliJ can be enabled with the following. It'll use our scalastyle_config.xml

> Intellij - You can enable scalastyle in Intellij by selecting Settings->Editor->Inspections, then searching for Scala style inspections.

screen shot 2016-05-26 at 10 05 56 am

And since `scalastyle-config.xml` is now in the path, IntelliJ will highlight the warnings (or error is configured for it).

screen shot 2016-05-26 at 11 17 41 am

PR #113, opened 2016-06-01
Scalastyle Checker: Part 2

jfalkner:
> WIP: not yet ready for merge.

Continues #95 and adds style checking for expected naming conventions. 

Adds naming style checks: Class, Object, Package, Method

Ignores a few easy-to-check items because it isn't clear from the style specs if this is needed. We can decide for our code base some other time.

- Enforce a single space before and after `+` symbols
- Enforce no whitespace at the end of lines

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -53,6 +53,7 @@
     <parameter name="illegalImports"><![CDATA[sun._,java.awt._]]></parameter>
    </parameters>
   </check>
+  -->
 <!-- Ignored - not limit to number of params in style specs, albeit less is normally preferred -->
 <!--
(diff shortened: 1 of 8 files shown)


jfalkner:
Ok, this wave is done and it is the last set of changes I have planned for the first pass. I left XML comments in scalastyle-config.xml to note the following.

- One rule is in the style guides but too much work to add now. Will make a GH issues for it
- A bunch of rules are easy to check but are arbitrary and not in the style guides. Will punt on these for now and wait for further group chat.

As-is, we have a decent start at checking for expected code style. I'll make a simple doc summarizing use.

jfalkner:
Tests pass locally.

PR #114, opened 2016-06-02
stress test that imports many datasets

skinner:
It is indeed hacky.


mpkocher:
Thanks, this is useful as a starting point. 

I'm going to take a quick pass at something in python that has concurrent access. I'll ping you and @natechols to get feedback, so we can converge on a model to use in the short term.

PR #115, opened 2016-06-02
Mk fixes

mpkocher:


mpkocher:
@skinner Does this work for you? I think the run design could be added pretty easily.

mpkocher:
This is still WIP, but it works ok for now.

PR #116, opened 2016-06-02
Code Style Doc in `projects/code_style.md`

jfalkner:
Added a simple code style doc to summarize the work and use. Includes noting common formatting issues observed in our codebase so that people can quickly see a valid format that'll pass the style checker.


mpkocher:
Can you move this to the root dir?

jfalkner:
Yeah, moving it now.

Issue #122, opened 2016-06-02
Diagnose Spray performance and better understand config


jfalkner:
Placeholder ticket for the work to help resolve the timeout issues @mpkocher highlighted at today's meeting. sqlite may be the culprit here, but we need to explain what is going on.

@skinner and @natechols are focusing on everything up through Slick. @jfalkner to work through the Spray side.

@mpkocher where is the stress test script and how to run?

jfalkner:
Is this the stress test? https://github.com/PacificBiosciences/smrtflow/pull/114

jfalkner:
Here is `stress.py` from @mpkocher. I think this is the stress test he mentioned on the call.

jfalkner:
Some notes about using the code from #115 and `stress.py`.

```
# already merged in master -- use it
git checkout master

# clean the old dir (optional)
make dataclean

# copy the data locally. must be on VPN
make test-data/smrtserver-testdata
```

And here is where it breaks. I think I need a different repo installed.

```
# duplicates the data a bunch, changing UUIDs
make test-data/copied-datasets
# tries to stress the server with uploads
make test-stress-mass-import
```

jfalkner:
PacificBiosciences/pbsmrtpip has the `dataset` command. Following the instructions in the [`python.md` README](https://github.com/PacificBiosciences/dev_readme/blob/master/python.md) gets you setup.

```bash
(ENV) jfalkner-mac:pbsmrtpipe jfalkner$ dataset
usage: dataset [-h] [--version] [--log-file LOG_FILE]
               [--log-level {DEBUG,INFO,WARNING,ERROR,CRITICAL} | --debug | --quiet | -v]
               [--strict] [--skipCounts]
...
```

jfalkner:
The stress test assumes that you have the `smrt-server-analysis` code running. Use the make target below.

```
# run analysis via make
make start-smrt-server-analysis
```
And in a diff shell, tmux, etc.
```
# run the stress test via pbsmrtpipe VENV
make test-stress-mass-import
```

jfalkner:
I'm successfully running the stress test and profiling the JVM. I'll dig in to the Spray docs and post more notes as interesting things are observed.

jfalkner:
[...]

PR #126, opened 2016-06-03
DB connection pooling for sqlite

jfalkner:
CC @mpkocher @skinner @smcclellan7 

This PR makes Flyway migrations work with SQLite and adds some other great features to our codebase. In all:

- Allows Flyway migrations to work with SQLite via [shared Connection hack](https://github.com/PacificBiosciences/smrtflow/pull/126/files#diff-fedccf5951a9b70358750de02eb32b86R41). See [notes in the new `SqliteAndFlywayUsageSpec` test](https://github.com/PacificBiosciences/smrtflow/pull/126/files#diff-2e2412b46c7f1ef39bae171cd49c6c86R11).
- [Explicit guard](https://github.com/PacificBiosciences/smrtflow/pull/126/files#diff-fedccf5951a9b70358750de02eb32b86R56) to ensure that any invalid `Connection` use in our codebase (similar to what caused SQLITE_BUSY errors) will be replaced with a `RuntimeException` that makes the root cause much more obvious.
- Database [connection pooling via DBCP](https://github.com/PacificBiosciences/smrtflow/pull/126/files#diff-fedccf5951a9b70358750de02eb32b86R43)], which allows for [`PreparedStatement` caching](https://github.com/PacificBiosciences/smrtflow/pull/126/files#diff-fedccf5951a9b70358750de02eb32b86R68) as well as [in-memory SQLite use for tests](https://github.com/PacificBiosciences/smrtflow/pull/126/files#diff-fedccf5951a9b70358750de02eb32b86R117.)

Our codebase should now also run faster -- beyond getting past the old locking problem!

1. The `PreparedStatement` caching should speed up Slick used as per [current profiling hotspot](https://github.com/PacificBiosciences/smrtflow/issues/122#issuecomment-223613724).
2. In-memory SQLite use for tests should be much faster than the old temp file use. No need to create and cleanup files.

## Tests

All tests are passing locally for me. See the new [`SqliteAndFlywayUsageSpec` test](https://github.com/PacificBiosciences/smrtflow/pull/126/files#diff-2e2412b46c7f1ef39bae171cd49c6c86R45) that guards against DB misuse and confirms expected `Connection` caching works for migrations.

```bash
sbt "smrt-server-analysis/testOnly SqliteAndFlywayUsageSpec"
...
[info] SqliteAndFlywayUsageSpec
[info] 
[info] Connection pooling for SQLite should
[info] + be in use
[info] + share Connection instances during Flyway migrations
[info] + guard against failing to close Connection instances
[info] + return unique Connection instances during normal, non-migration use
[info] 
[info] Total for specification SqliteAndFlywayUsageSpec
[info] Finished in 279 ms
[info] 4 examples, 0 failure, 0 error
[info] Passed: Total 4, Failed 0, Errors 0, Passed 4
[success] Total time: 7 s, completed Jun 7, 2016 4:39:49 PM
```


mpkocher:
This looks like a good path to pursue.  Can you comment out the flyway code path and try to get the stress test to run?

jfalkner:
Some links that were helpful

- [Slick database connection docs](http://slick.lightbend.com/doc/3.1.0/database.html). Notes use of DBCP and `DataSource`
- [DBCP config params docs](http://commons.apache.org/proper/commons-dbcp/configuration.html)
- [Latest Xerial sqlite JDBC driver](https://bitbucket.org/xerial/sqlite-jdbc/downloads). Copy version number to SBT dep.
- Related [article about DBCP with Scala](https://devcenter.heroku.com/articles/database-connection-pooling-with-scala#using-apache-dbcp)

jfalkner:
> This looks like a good path to pursue. Can you comment out the flyway code path and try to get the stress test to run?

Yep, on it. I think as of 2fd39a1 it should work. @skinner helped demistify this [Slick sqlite migration issue](http://stackoverflow.com/questions/30375308/how-to-avoid-flyway-migrations-locking-the-sqlite-database). Slick needs 2 Connection instances: one to lock schema and one to do the SQL. 

 I'm running the tests locally and also testing `PreparedStatement` caching. Hopefully all works.

jfalkner:
> I'd say flyway rather than slick

Correct, I should have noted Flyway not Slick.

jfalkner:
> Are connections thread-safe? Do we have to worry about concurrency when everything is using one connection?

We should be fine, unless something is directly trying to use JDBC outside the `Dal` instance. DB access is currently queued and handled by a single-threaded execution context. I don't believe it'll cause any threading issues as-is.

jfalkner:
Seeing this error. Sorting it out.

```bash
[error]      Could not unmarshal response to type 'com.pacbio.secondary.analysis.jobs.JobModels$EngineJob' for `responseAs` assertion: UnsupportedContentType(Expected 'application/json')
[error]      
[error]      Response was: HttpResponse(500 Internal Server Error,HttpEntity(text/plain; charset=UTF-8,There was an internal server error.),List(),HTTP/1.1) (JobExecutorSpec.scala:126)
```

jfalkner:
I merged the latest master. Still seeing the error above with `sbt "smrt-server-analysis/testOnly JobExecutorSpec"`. I'm working on sorting it out.

jfalkner:
@smcclellan7 I'm not entirely sure what was going on with the old `JobExecutorSpec` test. It had two sets of specs using `should` and it'd fail if both were checked but pass if either were checked individually. The issue seemed unrelated to my DB changes, albeit wasn't reproducible outside this branch. I split the spec in to two simpler to read ones plus an ABS.

jfalkner:
Tests all pass locally! 💪 😄 💪

jfalkner:
Testing out the stress.py test is up next. Thanks @skinner for advice on the full clean, setup and run commands.

```bash
make dataclean
make test-data/smrtserver-testdata
./stress.py
```

mpkocher:
Looks like this is almost there?

jfalkner:
Yep, two issues:

- [Failing CircleCI check](https://circleci.com/gh/PacificBiosciences/smrtflow/333) is a new error. I'm trying to replicate it locally.
- I ran `stress.py` and it still is having issues. Looks like no SQLITE_BUSY errors anymore, but I'd like to better confirm the issues are unrelated to this PR.

jfalkner:
@mpkocher @skinner here are the errors I'm seeing on the server side. I'll ask about these first two in Slack the last two I think I can trace and iron out on my own.

This one is thrown once or twice. Missing a config file? What is this and is a fix needed?
```bash
```

Failing `import pbreports` happens at least half a dozen times or so. I think this is one of the main reasons why `stress.py` doesn't fully pass. How do I get my VENV updated appropriately? 

```bash
[info] 2016-06-07  INFO [smrtlink-analysis-server-akka.actor.default-dispatcher-4] c.p.s.a.e.Python$ - Starting cmd List(python, -c, import pbreports)
[info] 2016-06-07  ERROR[Thread-5] c.p.s.a.e.Python$ - Traceback (most recent call last):
[info] 
[info] 2016-06-07  ERROR[Thread-5] c.p.s.a.e.Python$ -   File "<string>", line 1, in <module>
```

I think we have a race condition somewhere or code that isn't using the DAO in the typical way. I should be able to trace this down and fix it.
```bash
[info] 2016-06-07 20:55:52.465UTC INFO [smrtlink-analysis-server-akka.actor.default-dispatcher-23] c.p.s.a.e.a.EngineDaoActor - Unhandled engine DAO message. 'Failure(java.lang.RuntimeException: Can't have multiple sql connections open. An old connection may not have had close() invoked.)'
[info] 2016-06-07  WARN [ForkJoinPool-2-worker-15] c.p.c.a.BaseApi$ServiceActor - Non-fatal exception: 500 - Internal Server Error - Can't have multiple sql connections open. An old connection may not have had close() invoked.
[info] java.lang.RuntimeException: Can't have multiple sql connections open. An old connection may not have had close() invoked.
[info] 	at com.pacbio.secondary.smrtlink.actors.Dal$$anon$1.getConnection(JobsDao.scala:56)
[info] 	at slick.jdbc.DataSourceJdbcDataSource.createConnection(JdbcDataSource.scala:64)
[info] 	at slick.jdbc.JdbcBackend$BaseSession.conn$lzycompute(JdbcBackend.scala:415)
[info] 	at slick.jdbc.JdbcBackend$BaseSession.conn(JdbcBackend.scala:414)
[info] 	at slick.jdbc.JdbcBackend$SessionDef$class.prepareStatement(JdbcBackend.scala:297)
[info] 	at slick.jdbc.JdbcBackend$BaseSession.prepareStatement(JdbcBackend.scala:407)
[info] 	at slick.jdbc.StatementInvoker.results(StatementInvoker.scala:33)
[info] 	at slick.jdbc.StatementInvoker.iteratorTo(StatementInvoker.scala:22)
[info] 	at slick.jdbc.Invoker$class.foreach(Invoker.scala:48)
[info] 	at slick.jdbc.StatementInvoker.foreach(StatementInvoker.scala:16)
[info] 	at slick.jdbc.StreamingInvokerAction$class.run(StreamingInvokerAction.scala:22)
[info] 	at slick.driver.JdbcActionComponent$QueryActionExtensionMethodsImpl$$anon$1.run(JdbcActionComponent.scala:218)
[info] 	at slick.driver.JdbcActionComponent$QueryActionExtensionMethodsImpl$$anon$1.run(JdbcActionComponent.scala:218)
[info] 	at slick.backend.DatabaseComponent$DatabaseDef$$anon$2.liftedTree1$1(DatabaseComponent.scala:237)
[info] 	at slick.backend.DatabaseComponent$DatabaseDef$$anon$2.run(DatabaseComponent.scala:237)
[info] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[info] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[info] 	at java.lang.Thread.run(Thread.java:745)
```

Not sure about this one. Not an error but says it is?

```bash
[info] 2016-06-07  ERROR[ForkJoinPool-2-worker-7] c.p.s.s.a.JobsDao - Unable to update state of job id 7b3ec447-78de-4e14-ba60-b1ff63752af5

[info] 2016-06-07 20:55:52.251UTC ERROR[smrtlink-analysis-server-akka.actor.default-dispatcher-23] c.p.s.a.e.a.EngineManagerActor - Failed to update job state of CoreJob(7b3ec447-78de-4e14-ba60-b1ff63752af5,ImportDataSetOptions(/Users/jfalkner/tokeep/git/PacificBiosciences/smrtflow/test-output/dataset-f9f570fb-bde0-45c4-8e55-60a9fb8e3a1c/referenceset.xml,PacBio.DataSet.ReferenceSet)) with 7b3ec447-78de-4e14-ba60-b1ff63752af5
```

skinner:
> from pb-engine.pb-cmd-template Error No configuration setting found for key 'pb-engine.pb-cmd-template'

I actually don't know what causes this, but I've been seeing it for a long time and it doesn't seem to be causing problems AFAICT.  I've been assuming that it's okay; if it's not, it's way down the priority list.

> Failing `import pbreports`

I assume you installed pbreports?  Just in case that's the issue: we recently added a REQUIREMENTS.txt that has the necessary deps.

Going on the assumption that you do have pbreports installed in your venv:
For the stress test, we want to use the virtualenv for both the server process and the client process.  The server is the one trying and failing to import pbreports (the import failure is part of the server's check to see that it's installed).  When you launch the server, make sure you have the virtualenv activated (I do this on the command line; I'm not sure what you'd do if you're launching it from an IDE).

> I think we have a race condition somewhere
> `Can't have multiple sql connections open.`

I assume it's intermittent if you're thinking race condition?

I'm pretty sure we do have concurrent database use; I think the SQLITE_BUSY errors are evidence of that.  That's why we've been doing this stress testing, in order to reproduce that and verify the fix once we implement it.

Michael's original plan AIUI was to avoid concurrent database access (partly to avoid DB concurrency issues with sqlite).  So we were supposed to have only one actor accessing the database.  But we ended up with more than one; have you seen/heard the conversation about that?  I believe that's why we started seeing SQLITE_BUSY.

The database-accessing actors that I know of are JobsDaoActor (which uses JobsDao), RunServiceActor (which uses DatabaseRunDao), and EngineManagerActor (which also uses JobsDao).  AIUI, those actors can be scheduled to run concurrently on separate threads.  That's why I was asking about concurrency in an earlier comment on this PR.

There's work going on separately to address the concurrency issue; AIUI, Michael is working to consolidate those separate actors into one.  And Steve (in #140) is working on removing the actor layer from the database access, relying on slick 3.1 configuration to limit database access to a single thread.  I'm a little hazy on how Michael's work relates to Steve's work; they may have a belt-and-suspenders relationship.

> `Unable to update state of job id`

Not sure about this one; I've been lumping it in with "a whole bunch of stuff is broken since the slick 3.1 transition".

jfalkner:
Thanks @skinner. I'll ignore that  `pb-engine.pb-cmd-template` one. 

> I assume you installed pbreports? Just in case that's the issue: we recently added a REQUIREMENTS.txt that has the necessary deps.

I was using the wrong virtualenv. I originally setup `pbsmrtpipe`'s and had been using it. I see now that `smrtflow` has the REQUIREMENTS.txt. Switching to that.

> ... So we were supposed to have only one actor accessing the database. But we ended up with more than one; have you seen/heard the conversation about that? ...

Ah, thanks for the details above. I was not fully aware of the situation. Hmm, the different actors make this interesting. I think I can tweak the `Datasource` to have a non-blocking lock with a timeout that'll solve this problem for the codebase as-is. 

>>Unable to update state of job id

>Not sure about this one; I've been lumping it in with "a whole bunch of stuff is broken since the slick 3.1 transition".
SGTM, low priority for this one.

jfalkner:
Just hit an issue where matplotlib was complaining about no OSX backend. Fixed via [SO advice](http://stackoverflow.com/questions/21784641/installation-issue-with-matplotlib-python).

mpkocher:
```
# add to .bashrc
# http://matplotlib.org/faq/virtualenv_faq.html#pythonhome-function

function frameworkpython {
    if [[ ! -z "$VIRTUAL_ENV" ]]; then
        PYTHONHOME=$VIRTUAL_ENV /usr/local/bin/python "$@"
    else
        /usr/local/bin/python "$@"
    fi
}
```

mpkocher:
> There's work going on separately to address the concurrency issue; AIUI, Michael is working to consolidate those separate actors into one. And Steve (in #140) is working on removing the actor layer from the database access, relying on slick 3.1 configuration to limit database access to a single thread. I'm a little hazy on how Michael's work relates to Steve's work; Michael's may be the short-term plan, and Steve's the long-term.

The JobsDaoActor can not be entirely removed. It still needs all the job and engine related communication. For example, the dataset or project related message protocols could be deleted in favor of just accessing the DAO layer and returning a future. 

However, this would require that these locking issues are addressed and dao is accessible from spray using a sqlite backend.

jfalkner:
> There's work going on separately to address the concurrency issue

I'll revert cfe324d in favor of leaving this PR at dc93423. I think the code is good and the guard is what we want, as long as we address the concurrency issue elsewhere.

jfalkner:
It seems that `DBIO` is capable of opening multiple Connection instances and sometimes the actions don't close the DB connection. Use of `DBIO.seq` appears to do this and was causing DB issues when [`getJobStateByUUID` was invoked](https://github.com/PacificBiosciences/smrtflow/blob/f35789d2ae5e6408db70525e584e9a4c5808dfac/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala#L371-L374).

A fix for this is to generalize the `migrating` flag I was using for Flyway. Reuse the same Connection for each `db.run` invocation.

jfalkner:
Alright, as of 5c192a4 all of the cases where a shared `Connection` is needed appear to work. This generally fixes the `SQLITE_BUSY` error mode and enforces compatibility with SQLite's one active `Connection` requirement.

Running the tests locally. If they pass, I think we're good to merge here. It doesn't fully solve `stress.py` but it seems to be a big step toward improving DB access. I'd prefer to get this merged before it becomes an even bigger PR.

Related to this, I'll start a branch with a refactor noted in `Dal`. It is kind of a pain to debug DB access right now because of all the DBIO Future use and not knowing what thread queued what. The DB wrapper I added can clarify this  as well as dump profiling timing for logical database actions.

jfalkner:
Tests pass locally for me.

mpkocher:
Even if the tests aren't passing, I think we should merge this in. 

@jfalkner and @skinner Does that work for you?

jfalkner:
> Even if the tests aren't passing, I think we should merge this in.

> @jfalkner and @skinner Does that work for you?

Yep. Seems like a fair enough strategy as per chat on Slack too.

I'm still on the hook to help clarify and resolve the newer, seemingly less severe errors.

PR #130, opened 2016-06-06
import the original referenceset, import subreadset

skinner:
Import subreadset as part of the stress test.
Also, if the test referenceset had been imported with the original UUID, stress.py would run successfully, but if not, then then the dev_diagnostic pipeline would fail to run because it expected the original referenceset UUID.


jfalkner:
Thanks @skinner. I think this will help with the testing I'm trying to do in https://github.com/PacificBiosciences/smrtflow/pull/126

PR #131, opened 2016-06-06
added `pbservice run-pipeline` command

natechols:
With pbsmrtpipe-like syntax.  I had to modify the preset loader to get this to work.

### DIFF
--- a/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
+++ b/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
@@ -43,6 +43,7 @@ object Modes {
   case object IMPORT_FASTA extends Mode {val name = "import-fasta"}
   case object ANALYSIS extends Mode {val name = "run-analysis"}
   case object TEMPLATE extends Mode {val name = "emit-analysis-template"}
+  case object PIPELINE extends Mode {val name = "run-pipeline"}
   case object JOB extends Mode {val name = "get-job"}
   case object JOBS extends Mode {val name = "get-jobs"}
   case object DATASET extends Mode {val name = "get-dataset"}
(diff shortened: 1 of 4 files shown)


natechols:
@mpkocher you'd better take a look at my changes to PipelineTemplatePresetLoader.scala before we pull the trigger on this (no rush)

smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
```diff
@@ -537,6 +582,112 @@ class PbService (val sal: AnalysisServiceAccessLayer) extends LazyLogging {
...
+
+  protected def getPipelinePresets(presetXml: File): PipelineTemplatePreset = {
+    if (presetXml != null) {
+      val presetPath = Paths.get(presetXml.getAbsolutePath)
+      PipelineTemplatePresetLoader.loadFrom(presetPath)
+    } else defaultPresets
+  }
+
+  protected object OptionTypes {
+    val FLOAT = "pbsmrtpipe.option_types.float"
```
mpkocher: Is this enum defined somewhere else?

smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
```diff
@@ -217,13 +240,29 @@ class PbService (val sal: AnalysisServiceAccessLayer) extends LazyLogging {
   import AnalysisClientJsonProtocol._
 
   protected val TIMEOUT = 10 seconds
+  private lazy val entryPointsLookup = Map(
+    "PacBio.DataSet.SubreadSet" -> "eid_subread",
```
mpkocher: This believe this is defined in `com.pacbio.secondary.analysis.datasets.DataSetMetaTypes`

mpkocher:
This looks reasonable. Let merge and iterate.

smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
natechols: No, they are used here:

https://github.com/PacificBiosciences/smrtflow/blob/master/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/models/SmrtLinkJsonProtocols.scala#L41

and here:

https://github.com/PacificBiosciences/smrtflow/blob/master/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobs/SecondaryJsonProtocols.scala#L137

So it definitely needs to go somewhere central; how about here?

https://github.com/PacificBiosciences/smrtflow/blob/master/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobs/JobModels.scala

PR #132, opened 2016-06-07
make sure PB_SCALA_VERSION isn't null

skinner:


jfalkner:
Right on, made my morning. I think #126 may pass all tests now. Trying it out.

PR #133, opened 2016-06-07
use common smrtflow version code from smrt-common-models

skinner:
@mpkocher is this what you had in mind?  Since there's only one package/repo now, smrt-common-models seemed like the place for version code.  Please merge if it works for you.


mpkocher:
@jfalkner Does this work for you? 

I would suggest that the services and the commandline tools log the version (with the git sha) on startup.

PR #139, opened 2016-06-08
add run designs in stress test

skinner:
It may be a bit of a hack in the sense that it doesn't use pbservice.  But AFAICT pbservice doesn't do run designs, so this seemed like a reasonable approach.

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -8,3 +8,4 @@
 -e git://github.com/PacificBiosciences/pbcoretools.git#egg=pbcoretools
 -e git://github.com/PacificBiosciences/pbreports.git#egg=pbreports
 -e git://github.com/PacificBiosciences/pbsmrtpipe.git#egg=pbsmrtpipe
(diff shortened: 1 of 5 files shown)


smcclellan7:
LGTM

REQUIREMENTS.txt
```diff
@@ -8,3 +8,4 @@
 -e git://github.com/PacificBiosciences/pbcoretools.git#egg=pbcoretools
 -e git://github.com/PacificBiosciences/pbreports.git#egg=pbreports
 -e git://github.com/PacificBiosciences/pbsmrtpipe.git#egg=pbsmrtpipe
+requests
```
mpkocher: pbcommand has a dep on requests. This isn't necessary.

skinner: I was hesitant about relying on an indirect dep because I worried about the case where pbcommand drops its dep on requests.  Do you generally rely on indirect deps sticking around?

PR #140, opened 2016-06-08
Simplify Akka layer in smrt-server-base

smcclellan7:
- Only use akka actors for quartz scheduling in smrt-server-base (TODO: remove akka in s-s-link, s-s-analysis, s-s-analysis-internal)
- Remove "Dal" concept
- Use "Composer" model to gather objects requiring initialization from dependency injection
- Redesign health service with more flexible metrics (TODO: implement some useful metrics, e.g. latency, timeout rate, error rate, disk space)
- Remove db-backed health service


mpkocher:
Does the stress test pass with these changes?

mpkocher:
These changes need to be discussed.

PR #144, opened 2016-06-08
Migrate engine manager actor into JobDaoActor

mpkocher:
Still WIP


smrt-server-base/src/main/scala/com/pacbio/common/services/package.scala
```diff
@@ -150,7 +150,7 @@ package object services {
         } else {
           resp.message
         }
-        logger.warn(s"Non-fatal exception: ${resp.httpCode} - ${resp.errorType} - ${message}")
+        logger.warn(s"Non-fatal exception: ${resp.httpCode} - ${resp.errorType} - ${message} Request ${ctx.request}")
```
skinner: Awesome, I didn't realize the request was available here.  Do you know if this will log the body of a POST?  That would be super helpful for debugging that exception that was happening on the run design post.

mpkocher: It does. This should be helpful for debugging. Here's an example:

https://gist.github.com/mpkocher/0455ecc2b24b4c1930ca99bf644a8d2b

mpkocher:
I believe the test failures are due to the DB issues.

PR #150, opened 2016-06-10
implement models and basic functionality for new GmapReferenceSet

natechols:
I hadn't actually intended to modify all of these files...


smrt-analysis/src/test/scala/DataSetMergerSpec.scala
```diff
@@ -39,7 +39,7 @@ class DataSetMergerSpec extends Specification with LazyLogging {
       logger.info(s"Writing merged dataset to $p")
       DataSetWriter.writeHdfSubreadSet(mergedDataSet, p)
 
-      mergedDataSet.getVersion must beEqualTo("3.0.1")
```
mpkocher: For reasons that are unclear to me, the DataSet version should be 3.0.1.  This might be the "spec" of the file (?)

For the The SubreadSet, it has the ICS and PA version in the XML in 3.1.0.

natechols: So should I change this (and the VERSION constant) back to 3.0.1?

smrt-common-models/src/main/scala/com/pacbio/common/models/Constants.scala
```diff
@@ -6,7 +6,7 @@ import java.util.Properties
  * Created by mkocher on 10/13/15.
  */
 trait Constants {
-  final val DATASET_VERSION = "3.0.1"
+  final val DATASET_VERSION = "3.2.0"
```
mpkocher: This needs to be 3.0.1 

AK would have more details about what's going on here.

PR #155, opened 2016-06-13
DB Usage Refactor

jfalkner:
> Read for review. See below for the two PRs that need to merge in to this branch before this to master.

This is a refactor of our database access. Done to debug the DB issues but also to simplify our codebase and break out a logical chunk.

PRs to review that merge to this branch. They are split to simplify code review. You don't need to review anything but the `README.md` in this branch -- it is the only file unique to this branch.

- [ ] This PR: `db_refactoring` -- has the [README.md](https://github.com/PacificBiosciences/smrtflow/blob/db_refactor/smrt-server-database/README.md) and is a merge target for the below branches
- [x] #156 Refactor DB code to sub-project: `db_debugging`
  - Enforces that rest of the codebase can only used `run` to queue up DB access. Same as we had before, but now formally enforced.
  - Misc implementation details are now abstracted by `Database` and ensured to work correctly. e.g. migrations.
  - Wraps the DB access to allow unified control of connection pooling, commit/closing, shared connections, and adding any debugging or profiling.
  - [POC debugging and profiling](https://github.com/PacificBiosciences/smrtflow/tree/db_refactor/smrt-server-database#debugging-and-profiling) info to show where in our code the DB is used. (TODO: disable by default)
- [x] #157 Refactor to use the new `com.pacbio.database` package: `db_debugging_use_refactor`
  - Simplifies all the `dal.db.run` to just be `db.run`, and `Dal` is now called `Database`
  - Migration logic is now hidden and lazy performed abstracted. The rest of the codebase doesn't need to care about it. 
- [x] TBD a final PR that works in any changes from `master` so that this branch will be able to merge without conflict.

A related branch exists that demos H2 use so that we can confirm mechanistically that the rest of our code should work.

- Swap SQLite for H2. Show that multiple `Connection` support makes the code work as expected

PR #156, opened 2016-06-13
DB Usage Refactor: com.pacbio.database package

jfalkner:
See #155 for the merge target of this branch. This is split so that it highlights just the breakout of the `com.pacbio.database` package and the related `Database` class. The `SliteAndFlywayUsageSpec` has also been simplified and lives in this new package.

This branch won't pass tests outside of the `smrt-server-databse` sub-project. The greater test suite will pass for #155 (which is the merge target for this branch).


jfalkner:
e70d3bc is the merge of #157, which was passing tests. Merging this to #155 so we have the final PR.

PR #158, opened 2016-06-13
fasta-to-gmap-reference tool

natechols:


natechols:
Oops, this probably needs to write a .fai file as well.

PR #159, opened 2016-06-13
use task ID as report type ID

natechols:
@mpkocher alternatively, we could further split the task ID by '.' and take the last element, so 'mapping_stats', 'ccs_report', and so on


mpkocher:
I think I would like to take second pass at this, but this might be fine for now.

Issue #161, opened 2016-06-14
Resolve `stress.py` errors to confirm SQLite use is fixed


jfalkner:
CC @skinner 

Followup work to #156. It appears SQLite use now works as expected on the server-side. No locking and the profiling of DB use shows all queries succeeding -- no errors. 

Yet `stress.py` still has issues. The work here is to sort out what is going on with the test. Hopefully this is  mostly a fix on the test ENV or in the test itself.

The setup here is to have one terminal with the server running.

```
# activate the virtualenv
source ENV/bin/activate

# clear out old DB data
rm -fr smrt-server-analysis/db

# run the server
make start-smrt-server-analysis > debug_db.txt
```

And one running the stress test.
```
# activate the virtualenv
source ENV/bin/activate

# setup needed data
make dataclean
make test-data/smrtserver-testdata

# run the stress test
(ENV)$ ./stress.py 2> stress.err > stress.out
```

Here are example logs of the issues I'm seeing.

[stress_err.txt](https://github.com/PacificBiosciences/smrtflow/files/314583/stress_err.txt)
[stress_out.txt](https://github.com/PacificBiosciences/smrtflow/files/314584/stress_out.txt)

jfalkner:
I reran everything to get the latest logs at default logging level. 

`make start-smrt-server-analysis 2> server_err.txt > server_out.txt`

[server_err.txt](https://github.com/PacificBiosciences/smrtflow/files/314651/server_err.txt)
[server_out.txt](https://github.com/PacificBiosciences/smrtflow/files/314652/server_out.txt)

`./stress.py 2> stress.err > stress.out`

[stress_err.txt](https://github.com/PacificBiosciences/smrtflow/files/314653/stress_err.txt)
[stress_out.txt](https://github.com/PacificBiosciences/smrtflow/files/314654/stress_out.txt)

jfalkner:
I've added some tooling in [`http_debugging`](https://github.com/PacificBiosciences/smrtflow/tree/http_debugging) to clarify what code is exercised on the server-side. Along with testing `stress.py` I'm going to see if removing the actor layer helps with this logic. I think the tooling can be generalized to something that logs/debugs all HTTP routes.

Based on the existing `server_out.txt` above the set of URLs can be pulled with:

```
# awk that shows all URL endpoints based on logging from our `RoutedHttpService` that is used by `ServiceActor`
awk '/http/{ print $0 }' server_out.txt
```

[...]

PR #170, opened 2016-06-16
Exposes `.transactionally` and other incompatible DB use bugs

jfalkner:
A few quick updates to expose Slick use that appears to be fundamentally incompatible with SQLite. Two bad assumptions are fixed here:

1. DB use wouldn't hard timeout. It now will after 10s
2. Sharing the `Connection` instance is now limited back to just Flyway migrations. It lead to cases where an SQL transaction could attempt to be created in an existing transaction, which locks up SQLite.

One thing that appeared to lock up an `Actor` was use of Slick's `.transactionally`. I'm still sorting out if we're limited to not using this at all or just being aware that it can potentially be an issue.


jfalkner:
This makes the stress test work a lot better on the current `master`. See #161

PR #171, opened 2016-06-16
add indexes to sqlite tables

skinner:


skinner:
@smcclellan7 I recall there being a raw SQL migration that wasn't working earlier; this one seems to work for me but I haven't done a comparison.  Does the `db.run(DBIO.seq`... look right to you?

Also, I didn't see a nice way to do this without using raw SQL, but if you have ideas about that they'd be welcome.

mpkocher:
@smcclellan7 can you review and merge?

skinner:
@smcclellan7 test mention

PR #173, opened 2016-06-17
Akka streamline

smcclellan7:
- Remove deprecated StatusServiceActor
- Streamline CleanupServiceActor. (Only keep messages required for quartz scheduling.)


smrt-server-base/src/main/scala/com/pacbio/common/actors/CleanupDao.scala
```diff
@@ -64,17 +65,17 @@ abstract class AbstractCleanupDao(clock: Clock, system: ActorSystem, loggerFacto
   // TODO(smcclellan): Remove this?
   def loadJobs(): Unit // Abstract method to load jobs from storage into memory on initialization
 
-  override def getAllJobs(): Set[CleanupJobResponse] = allJobs.values.map(_.toResponse).toSet
+  override def getAllJobs: Future[Set[CleanupJobResponse]] = Future(allJobs.values.map(_.toResponse).toSet)
 
-  override def getJob(id: String): CleanupJobResponse = {
+  override def getJob(id: String): Future[CleanupJobResponse] = Future {
```
mpkocher: I believe this in-memory Dao will need to use synchronized?

smcclellan7: Done.

PR #174, opened 2016-06-17
modified package-info.java to include namespace definitions

amaster08:


mpkocher:
We'll need a model to regenerate the classes and not overwrite these files.

PR #175, opened 2016-06-17
New 'metrics' model for health service

smcclellan7:


mpkocher:
Why did you change the API ?

smcclellan7:
I think this will be much more useful.  The way the old API worked, the client had to set the level of the health gauge explicitly with each update.  That would work fine for something like, say, disk space, where you don't need to coordinate info from multiple sources.  But let's say you wanted to track something like the latency of all the smrtlink service endpoints.  You go to each of the service endpoints and put a stopwatch on it, but then what?  You can't know how to set the health of the ALL_ENDPOINTS_LATENCY metric from just that one data point.

This would solve that problem.  Instead of setting the health level with every update, you just send the latency, and you set up the metric to take, say, the average latency over the last hour, and you define your health levels to be like, > 0.5 sec =  caution, > 1 sec = alert, > 3 sec = critical.  Or whatever.  But the point is, the health service can collate the data for you.

Also, with this new API, updates can be tracked by multiple metrics.  So lets say you had your ALL_ENDPOINTS_LATENCY metric and a JOBS_MANAGER_SERVICE_LATENCY metric and a POST_DATA_LATENCY metric.  When the JobsManagerService gets a POST request, instead of sending updates to three different metrics, you send one update with three tags, and all three metrics pick up that update.

I think it's a good direction to move in, in general, and I also think it could be useful for tracking some of these timeouts and sqlite_busy errors and such.

mpkocher:
There has be several discussion with ICS and PA about having a general Alarm/Health model. I believe you were in some of these meetings. 

The models that are currently proposed are different from these. These models will probably require substantial changes. 

Also for general profiling, I would prefer to use http://kamon.io/ or similar. 

The Health/metrics are designed to be system specific metrics, such as disk space is full, or the last N analysis (or import) jobs per time has exceeded a threshold. 

Nevertheless, let's just merge and limit the usage of this until we're synced with ICS and PA.

PR #182, opened 2016-06-20
H2 as the embedded database, replacing SQLite

jfalkner:
> WIP: Still getting tests to pass

Refactor to use H2 as the embedded database instead of SQLite. Don't merge this PR until we've had more discussion about the transition. I'm setting it up so that normal dev work is fine via H2, but we need some way to port existing SQLite data to H2 (or add and support Postgres) before this is in master.

Stuff left TODO:
- [x] Change JDBC driver use and remove the SQLite specific shared connection
- [ ] Get all the tests working
- [ ] Finalize the transition plan from SQLite to whatever is next

For the most part, this PR currently shows how to drop in H2 support in our codebase and it may be a helpful branch for testing if issues are SQLite specific or may have a root cause elsewhere.


smrt-server-analysis/src/main/resources/application.conf
```diff
@@ -25,7 +25,7 @@ akka {
 
 pb-services {
   # This can be provided as just a path or with jdbc:sqlite: prefix
-  db-uri = "jdbc:sqlite:db/analysis_services.db"
+  db-uri = "jdbc:h2:/tmp/analysis_services.db"
```
jfalkner: H2 doesn't support the notion of relative paths. Worth noting since we don't want to use `/tmp` for a file that should persist in a prod environment, and because our old SQLite JDBC URL was relative path based.

smrt-server-database/src/main/scala/com/pacbio/database/Database.scala
```diff
@@ -29,13 +23,7 @@ import ExecutionContext.Implicits.global
  *   <li>Lazy-initialization of required DB directory (SQLite specific)</li>
  *   <li>Lazy-migration via Flyway (upon first use of DB)</li>
  *   <li>Connection pooling via DBCP, including caching PreparedStatements</li>
- *   <li>Flag for enabling Connection sharing in required cases. Required for SQLite since one
- *   Connection is available for use but the code sometimes requires multiple</li>
- *   <ul>
- *     <li>Flyway migrations lock the DB schema (one connection) then migration (second connection).</li>
- *     <li>Slick <code>DBIO.seq</code> use will open multiple connections</li>
- *   </ul>
```
jfalkner: Lots of funny business required to make SQLite work is dropped from this file. See comments above and red lines below.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
```diff
@@ -45,7 +45,11 @@ trait SmrtLinkDalProvider extends DalProvider {
 trait TestDalProvider extends DalProvider {
   override val db: Singleton[Database] = Singleton(() => {
     // in-memory DB for tests
-    new Database(dbURI = "jdbc:sqlite:")
+    new Database(dbURI = "jdbc:h2:mem:")
+    // see http://h2database.com/html/features.html#in_memory_databases
+    val dbid = UUID.randomUUID()
+    println("JDBC Test JDBC URI: "+s"jdbc:h2:mem:$dbid;DB_CLOSE_DELAY=3")
+    new Database(dbURI = s"jdbc:h2:mem:$dbid;DB_CLOSE_DELAY=3")
```
jfalkner: I think this random UUID can be dropped. I'll do it after tests are working with the named in-mem DB and it is easy to confirm.

smrt-server-analysis/src/main/resources/application.conf
skinner: from [the h2 features page](http://www.h2database.com/html/features.html):
> If no or only a relative path is used, then the current working directory is used as a starting point.

Sounds like it supports relative paths.

jfalkner:
The bulk of the tests are now working. There are two issues to raise:

Biggest issue. I've yet to figure out how to alternate the correct Slick driver depending on the database, meaning I'm not sure we can easily do multi-database support. e.g. `import SQLiteDriver.api._` vs `import H2Driver.api._`

Anyone know how to resolve this error? Looks like a column that the migrations aren't setting up correctly.

```
Caused by org.h2.jdbc.JdbcSQLException: Column "state" not found; SQL statement:
[error] insert into "projects" ("name","description","state","created_at","updated_at")  values (?,?,?,?,?) [42122-192]
```

smcclellan7:
Would something like this work?

`
import com.pacbio.database.Database

trait JdbcComponent {
  val driver: JdbcDriver
}

trait SqliteComponent extends JdbcComponent {
  import slick.driver.SQLiteDriver
  override val driver = SQLiteDriver
}

trait H2Component extends JdbcComponent {
  import slick.driver.H2Driver
  override val driver = H2Driver
}

abstract class MyDatabaseWidget(db: Database) {
  this: JdbcComponent =>
  import driver.api._

  // ... do database stuff ...
}

class H2DatabaseWidget(db: Database) extends MyDatabaseWidget(db) with H2Component
class SqliteDatabaseWidget(db: Database) extends MyDatabaseWidget(db) with SqliteComponent
`

jfalkner:
> Would something like this work?

Will that work in `TableModels.scala`? It isn't a widget. Could try it to find out.

We end up doing that db-specific import in many places in the codebase.

smcclellan7:
I think you could do TableModels similarly.

```
abstract class TableModels {
  this: JdbcComponent =>

  // ... define table schemas ...
}

object H2TableModels extends TableModels with H2Component
object SqliteTableModels extends TableModels with SqliteComponent
```

It is a little unfortunate to have to do this with every class definition that requires a `import slick.driver.{SQLiteDriver|H2}.api._`, but I don't see a better way.  :-P

jfalkner:
> I think you could do TableModels similarly.

@smcclellan7 how will the abstract class compile? My understanding is that it'll fail to compile unless the import is present due to some of the DB-specific stuff that gets imported but has the same names.

smcclellan7:
@jfalkner You're probably right; I'm just spit-balling. You might end up having to make two full copies of the TableModels class.

jfalkner:
Ugh, hitting a wall of sorts here. As-is, [H2 works in this branch with Slick](https://github.com/PacificBiosciences/smrtflow/blob/h2/smrt-server-database/src/test/scala/com/pacbio/database/H2SlickUsageSpec.scala) but I'm struggling to get our migrations working with it. 

Two fronts where I'm stuck.
- I tried copying the old Slick migrations. They run but then the app complains about missing columns. Unclear to me if the changes commit or not. Unclear to me if our custom SlickMigration trait is working as expected with the current H2 setup.
- I manually dumped the SQL from SQLite and converted it to be H2 compatible. Deleted the migration files then ran it using a JDBC driver param. Stuck on seeing "table already exists" style messages. Noted the steps below.

Converting the SQL had one main kink in it where H2 appears to not support indexes on BLOB fields, which we do use in SQLite. This means some primary key and FK indexes were dropped.

1. You can manually load the SQL script with a JDBC driver command such as: `"jdbc:h2:/tmp/analysis_services;INIT=runscript from '/Users/jfalkner/tokeep/git/PacificBiosciences/smrtflow/smrt-server-link/src/main/resources/db_todelete/V1__initial_schema.sql';"`
2. Switch from SQLite's function for string-formatting the current timesamp to H2's: `FORMATDATETIME('yyyy-MM-dd HH:mm:ss',CURRENT_TIMESTAMP())`
3. s/AUTOINCREMENT/AUTO_INCREMENT/
4. INT range doesn’t match for timestamps. [SQL auto-sizes](https://www.sqlite.org/datatype3.html) and [H2 needs explicit](http://www.h2database.com/html/datatypes.html#int_type).
  - s/"created_at" INTEGER/"created_at" BIGINT/
  - s/"updated_at" INTEGER/"updated_at" BIGINT/
5. Reorder CREATE statements to FK match up.
  - engine_jobs above job_tags
  - projects above project_users
6. Can’t unique/index UNIQUE_ID BLOB column
  - Drop PRIMARY KEY `s/BLOB PRIMARY KEY/BLOB/`
  - Drop FK constraints "DM_SUMMARY_FK" and "RUN_FK"
7. Make FK constraints have unique names.
  - "SUMMARY_FK" had two uses made "DM_SUMMARY_FK" and “RDM_SUMMARY_FK"
  - "job_fk" was used twice. Now is “je_job_fk” “jt_job_fk”
8. Added quotes to all names. H2 thinks the table and var names include quotes. Should prb drop the quotes in the CREATE statements.

jfalkner:
@mpkocher advice on how to proceed here? I'm inclined to pause H2 work or hand it off to someone who wants to pick up the torch. SQLite is passing `stress.py` for me locally, and I'm starting to think that it is a crutch we'll need to lean on longer. Improving the stress.py to be all Scala and ironing out any lingering issues may be the practical route. I'd also like to page back in to some internal server work.

Big chunks of work I'm not sure make sense to do now.

- Refactoring our codebase to support multiple databases. Despite using Slick, this isn't trivial. e.g. import discussion above.
- Getting migrations to work in H2. Seemed simple, but is becoming a time sink.
- Migration tool. I have no progress on this one. It needs attention before we'd be ready to roll it out.

smcclellan7:
I'd be happy to take a swing at the H2 migrations.  I doubt I'll magically have any insight that you didn't already have, but who knows.  :-P

jfalkner:
@smcclellan7 Thanks. Hopefully you have some luck. It took me awhile to figure out how the Slick-based migrations were working. You'll have fresh eyes and more experience with that code.

mpkocher:
My suggestion to this would be to bypass the migrations entirely and build the tables directly from the ddls. 

Then get stress.py to work. If that works, then delete the extra sqlite specific database layer and try to get the stress.py to work.

jfalkner:
The SQLite specific data layer is already deleted in this branch. If the branch is working, we'll have to add back in SQLite support.

smcclellan7:
Quick question, do we want to be able to preserve any data that's in the user's SQLite db (if they have one) and migrate that data to H2?  Or are we okay with just building an empty H2 database from scratch?

If it's the former, we have to run all of the current flyway migration steps, move the data from SQLite to H2, and then start a new chain of H2 flyway migrations to operate on the new database.  It's a little tricky, but I think it can be done with custom MigrationResolvers.

https://flywaydb.org/documentation/api/javadoc/org/flywaydb/core/api/resolver/MigrationResolver

jfalkner:
@smcclellan7 I just scanned through my old PRs and branches to close out and clean up all the stale stuff pre-BitBucket era. I'll assign this PR your way. All yours to close out appropriately.

Issue #40, opened 2016-05-13
Implement Idable for Job/Dataset Id/UUID 


natechols:
The main place this is a problem so far is in the client API and particularly the 'pbservice' replacement - the `PbSmrtPipeServiceOptions` model contains `BoundServiceEntryPoint` where datasetId is `Int`, but we need to be able to handle either Int or UUID in the JSON config files.

mpkocher:
This should use a model similar to this:

https://gist.github.com/mpkocher/8d37114680b9c39d9adb119f74ce76e1

mpkocher:
The IdAble model is in common models subproject. It's adopted in a few places. It will be a continuing effort to clean this up.

PR #183, opened 2016-06-21
workaround for ID vs. UUID inconsistency in entry point JSON

natechols:
The change to ValidateImportDataSetUtils.scala is a bit hacky - I was trying to avoid changing anything relating to the database


mpkocher:
A longer term solution is tracked in filed under #40

PR #189, opened 2016-06-22
add --non-local import for import-dataset command

### LINKED ISSUE #178: Add --non-local to 'pbservice import-dataset'
Currently, importing datasets can only be done with access to the FS where the dataset exists.

Expand importing to include non-local (to where pbservice is running) importing of datasets would require:

- explicitly providing the DataSet metatype
- skipping any validation
- additional changes to the ServiceAccessLayer (TBD)

natechols:
Fixes #178

### DIFF
--- a/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
+++ b/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
@@ -95,6 +95,7 @@ object PbServiceParser {
       ploidy: String = "",
       maxItems: Int = 25,
       datasetType: String = "subreads",
+      nonLocal: String = null,
       asJson: Boolean = false,
       pipelineId: String = "",
       jobTitle: String = "",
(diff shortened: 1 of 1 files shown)


natechols:
This is difficult to test, but it behaves as I expected.  I'm not sure if `--non-local <MetaType>` is necessarily the most obvious argument syntax though.

smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
```diff
@@ -545,8 +549,11 @@ class PbService (val sal: AnalysisServiceAccessLayer,
     ).toArray ++ f.listFiles.filter(_.isDirectory).flatMap(listDataSetFiles)
   }
 
-  def runImportDataSets(f: File): Int = {
-    if (f.isDirectory) {
+  def runImportDataSets(f: File, nonLocal: String = null): Int = {
```
mpkocher: Can this be changed to Option[String] ?

PR #192, opened 2016-06-22
Stress test makefile rule

skinner:


makefile
```diff
@@ -52,3 +53,23 @@ validate-pipeline-view-rules:
...
+	    RUN=run-$$(date +%s) && \
+	    RUNDIR=test-output/stress-runs && \
+	    OUTDIR=$$RUNDIR/$$RUN && \
+	    mkdir -p $$OUTDIR && \
+	    rm -f $$RUNDIR/latest && \
+	    ln -s $$RUN $$RUNDIR/latest && \
+	    rm -f smrt-server-analysis/db/analysis_services.db; \
+	    rm -rf smrt-server-analysis/jobs-root/*; \
+	    sbt smrt-server-analysis/compile && \
+	    SERVERPID=$$(bash -i -c "sbt -no-colors \"smrt-server-analysis/run --log-file $(CURDIR)/$$OUTDIR/secondary-smrt-server.log\" > $$OUTDIR/smrt-server-analysis.out 2> $$OUTDIR/smrt-server-analysis.err & echo \$$!") && \
```
mpkocher: Would it be useful to call the jar file directly (this is what is done in the prod env) versus adding this extra sbt layer?

skinner: I was thinking about the use case where you're changing the code and want to see the stress test effect.  I guess we could run `sbt smrt-server-analysis/assembly` every time; it's a bit time consuming though.  Does running from the jar buy us much?

PR #196, opened 2016-06-23
database and services support for GmapReferenceSet

natechols:
I think this works - enough that I can import a dataset and see it in `/secondary-analysis/datasets/gmapreferences`.  Are we unit-testing stuff like this or leaving it to integration tests?


natechols:
@skinner please review the migration, I'm not 100% certain about this

mpkocher:
@skinner can you review this? Looks good to me.

PR #195, opened 2016-06-22
Guard for nested queries

jfalkner:
> WIP: need to flesh out docs and improve this description. Good to merge this PR when CircleCI is green. This code doesn't fix `stress.py`, rather it just shows how to see what DB calls need un-nesting. `unwrap_nested_db_run_usage` is the branch with the example fixes.

CC @mpkocher @skinner 

This includes some DB debugging improvements that I used to help clarify the nested query issue that appears to be the root cause of #161.

Running this branch should trigger the new guard that shows a more informative error such as the following, which notes that `insertDataStoreByJob` at `JobsDao.scala:505` timed out probably due to nested queries.

```
[info] 2016-06-22  ERROR[ForkJoinPool-2-worker-1] c.p.d.LoggingListener - [PacBio:Database]  RDMS timeout for com.pacbio.secondary.smrtlink.actors.DataSetStore$class.insertDataStoreByJob(JobsDao.scala:505)
[info] java.lang.Exception: Nested db.run() calls?
[info] 	at com.pacbio.database.Database$$anonfun$run$2$$anonfun$apply$3$$anonfun$apply$mcV$sp$6.apply(Database.scala:200)
[info] 	at com.pacbio.database.Database$$anonfun$run$2$$anonfun$apply$3$$anonfun$apply$mcV$sp$6.apply(Database.scala:200)
[info] 	at scala.collection.immutable.List.foreach(List.scala:381)
[info] 	at com.pacbio.database.Database$$anonfun$run$2$$anonfun$apply$3.apply$mcV$sp(Database.scala:200)
[info] 	at com.pacbio.database.Database$$anonfun$run$2$$anonfun$apply$3.apply(Database.scala:200)
[info] 	at com.pacbio.database.Database$$anonfun$run$2$$anonfun$apply$3.apply(Database.scala:200)
[info] 	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
[info] 	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
[info] 	at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
[info] 	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
[info] 	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
[...]

PR #197, opened 2016-06-23
Fixes `stress.py` in master, regardless of # --nprocess specified

### LINKED ISSUE #161: Resolve `stress.py` errors to confirm SQLite use is fixed

jfalkner:
CC @mpkocher @skinner 

Fixes #161

This PR should be the last set of changes to make `master` pass the existing `stress.py` integration test using SQLite, without the limitation of the `--nprocesses 1` setting. Previous DB refactoring #146 #170 #195

>NOTE! This PR only fixes current `stress.py` testing. I bet we have a few other cases where nested queries are being used and we need to do similar fixes. Those can be fixed in subsequent PRs.

The main fix here is un-nesting `db.run()` use so that one query won't block on another. It seems we simply can't use this practice reliably as long as we're using SQLite and limited to one connection.

GOOD: It'll work
- `DBIO.seq` and `DBIO.sequential` these are both fine as long as the list of statements doesn't invoke new `db.run()`
- `transactionally` is fine to use 

BAD: It'll break
- Don't ever have a `db.run()` `DBIOAction` that relies on a nested `db.run()` invocation
  - e.g. No way to have multiple connections with transactions -- only one transaction may be going at a time. If a parent query uses `.transactionally` it starts a transaction on the only `Connection` and the nested query's transaction will hopelessly block and timeout.
  - e.g. No way to reliably nest `db.run()` even if the nested queries are read-only. You must run the nested queries first then the one that uses the values. If you compose them all together, then there is a race condition where the parent query may block on a nested query that'll never execute due to the single DB connection that runs queries in serial.

See #195 for a [README.md](https://github.com/PacificBiosciences/smrtflow/tree/master/smrt-server-database#detecting-nested-queries-and-sqlite) link and some more details on how to debug cases that'll break.

## Testing

Running the test (normal `sbt test` passes too).
```
# launch the server
$ rm -fr smrt-server-analysis/db
$ make start-smrt-server-analysis > server_out.txt 2> server_err.txt

# run stress.py (in a different term)
./stress.py 2> stress_err.txt > stress_out.txt
```

And see that `stress_out.txt` succeeds. Would be nice to have a more obvious final `PASSED` status.

```
# Look for final summary of stress.py work.
$ cat stress_ou.txt
...
Dataset Summary
----------
SubreadSets 5
HdfSubreadSets 0
ReferenceSets 6
AlignmentSets 0
----------
All Job types Summary
----------
import-dataset Jobs 11
State SUCCESSFUL 11
----------
merge-datasets Jobs 0
----------
convert-fasta-reference Jobs 0
----------
pbsmrtpipe Jobs 3
State RUNNING 1
State CREATED 2
```

And no errors.

```
# no stress.py errors
awk '/ERROR/ || /Error/ || /imeout/' stress_out.txt

# no server_out.txt errors
awk '!/info/ && ( /ERROR/ || /Error/ || /imeout/ )' server_out.txt 

# no server_err.txt errors
cat server_err.txt
```


jfalkner:
@mpkocher @skinner @smcclellan7 Ok, about as good as I could get with the first pass. Any advice or improvements would be helpful. I'm new to Slick and Akka.

I'll double-check that tests pass and that `stress.py` still passes, then I think we're good.

jfalkner:
Tests pass locally for me.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
```diff
@@ -562,6 +570,10 @@ trait DataSetStore extends DataStoreComponent with LazyLogging {
   private def getDataSetMetaDataSet(uuid: UUID): Future[Option[DataSetMetaDataSet]] =
     db.run(dsMetaData2.filter(_.uuid === uuid).result.headOption)
 
+  // removes a query that seemed like it was potentially nested based on race condition with executor
+  private def getDataSetMetaDataSetBlocking(uuid: UUID): Option[DataSetMetaDataSet] =
+    Await.result(getDataSetMetaDataSet(uuid), 23456 milliseconds)
```
jfalkner: After all the work, I think this may be the block that makes it all work. Before these changes it seemed to make nested queries when calls to [insertDataSet](https://github.com/PacificBiosciences/smrtflow/pull/197/files#diff-fedccf5951a9b70358750de02eb32b86R520) were made

jfalkner:
Merging since we're going from broken `stress.py` to usable. Will address any cleanup as needed.

PR #198, opened 2016-06-23
models for new testkit config

natechols:
This is only the first step, but I'd like to get some quick feedback before using this.  Currently the JSON spec looks something like this:

```
{
  "jobName": "test_job",
  "workflowXml": null,
  "entryPoints": [{
    "entryId": "eid_subread",
    "path": "/path/to/subreadset.xml"
  }, {
    "entryId": "eid_ref_dataset",
    "path": "/path/to/referenceset.xml"
  }],
  "description": "simple test config",
  "jobType": "pbsmrtpipe",
  "outputDir": null,
  "presetXml": "preset.xml",
  "reportTests": [{
    "reportId": "example_report",
    "rules": [{
      "attrId": "mapped_reads_n",
      "op": "gt",
      "value": 100
    }, {
      "attrId": "concordance",
      "op": "ge",
      "value": 0.85
    }, {
      "attrId": "instrument",
      "value": "54006"
    }]
  }],
  "pipelineId": "pbsmrtpipe.pipelines.sa3_sat"
}
```

In the old testkit code, the report attribute rules are specified like this:

  "mapped_reads_n__gt": 100

I find this syntax easier to work with, and it's easily converted into the ReportAttributeRule objects, but I'm not sure how good an idea it is to have the JSON structure diverge from the underlying data model.  Do we care?

### DIFF
--- a/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/testkit/Models.scala
+++ b/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/testkit/Models.scala
@@ -37,7 +37,6 @@ object TestkitModels {
     pipelineId: Option[String],
     workflowXml: Option[String],
     presetXml: Option[String],
-    outputDir: Option[String], // XXX unused here
     entryPoints: Seq[EntryPointPath],
     reportTests: Seq[ReportTestRules])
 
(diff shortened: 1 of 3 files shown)


natechols:
PS. Note that jobType could just as easily be "import-dataset" or "import-fasta", in which case several other optional fields would be None - but report tests would work just as well.

mpkocher:
Looks good. 

Can you clarify why `outputDir` is in there?

natechols:
> Can you clarify why outputDir is in there?

Because the goal (I think) is to be able to use the same JSON files for pbtestkit-runner in pbsmrtpipe - I don't want to have to maintain parallel configurations for testing on services vs. the command line.  Of course this would require additional data items to specify the Python test modules, but I was hoping that `spray.json` would just ignore them.

natechols:
That said - `outputDir` would not be used in the Scala testkit, since it's doing everything through services, so maybe I should just remove that?

natechols:
I'm going to make the test a little more thorough...

PR #202, opened 2016-06-24
Eager Flyway migrations

jfalkner:
CC @skinner 

Runs Flyway migrations in an eager fashion but otherwise keeps the same guard in place to make sure migrations have run before other `db.run` use occurs.

This is a followup to a discussion on Slack where @skinner noted that some timeouts were being observed while running `stress.py` because the first few `db.run` usages blocked on relative long-running migrations.


jfalkner:
Once the server starts up, migrations should now be done.

jfalkner:
Tests pass locally for me.

Issue #205, opened 2016-06-24
`stress.py` still failing. No server-side timeouts, errors


jfalkner:
@skinner @mpkocher 

With today's code I'm no longer seeing `stress.py` pass. It sporadically fails. There appear to be no errors no timeouts in the logs. Instead I think it may be some sort of race condition where the status check occurs before the DB records the import.

Here is what I'm seeing in the logs for the failed import. After data import a GET is issued to ` http://localhost:8070/secondary-analysis/datasets/<dataset>`. When manually checked after the test, this URL shows the right result.

During the test, the one UUID that failed had this in the logs.

```
##$ cat server_out.txt | grep "1a569504-1c3c-e295-3af3-9a794140de12"
...
# HTTP GET is logged (response is not -- see next)
[info] 2016-06-24 18:28:54.421UTC INFO [smrtlink-analysis-server-akka.actor.default-dispatcher-16] a.a.RepointableActorRef - request: HttpRequest(GET,http://localhost:8070/secondary-analysis/datasets/1a569504-1c3c-e295-3af3-9a794140de12,List(Host: localhost:8070, Connection: keep-alive, Accept-Encoding: gzip, deflate, Accept: */*, User-Agent: python-requests/2.10.0, Content-Type: application/json),Empty,HTTP/1.1)

# 404 returned because import isn't done
[info] 2016-06-24  WARN [ForkJoinPool-2-worker-11] c.p.c.a.BaseApi$ServiceActor - Non-fatal exception: 404 - Not Found - Unable to find dataset 1a569504-1c3c-e295-3af3-9a794140de12 Request HttpRequest(GET,http://localhost:8070/secondary-analysis/datasets/1a569504-1c3c-e295-3af3-9a794140de12,List(Host: localhost:8070, Connection: keep-alive, Accept-Encoding: gzip, deflate, User-Agent: python-requests/2.10.0, Content-Type: application/json),Empty,HTTP/1.1)

# import ends after last attempted GET
[info] 2016-06-24  INFO [ForkJoinPool-2-worker-15] c.p.s.s.a.JobsDao - imported ReferencecSet 1a569504-1c3c-e295-3af3-9a794140de12 from /Users/jfalkner/tokeep/git/PacificBiosciences/smrtflow/test-output/dataset-159f0cd7-636d-4cb8-9d02-b8ce183ce251/referenceset.xml
```

Whereas all others that passed have a GET occur after the import succeeds.

```
# example of a successful import
[...]

PR #210, opened 2016-06-29
have JobExecutor wait for imports to finish

skinner:
@jfalkner @mpkocher

With this change, I no longer see import failures with stress.py.  I still see stress.py failures occasionally on analysis jobs, but the failure rate went from 2-6 failures per run to one failure every few runs.

I started with a smaller change that just `map`ped the futures from `ImportDataStoreFile` into the future returned by `importDataStore`.  But that resulted in a ClassCastException.  It turned out that the `.mapTo[Either[FailedMessage, SuccessMessage]]` in `importDataStoreFile` wasn't right.  I imagine the type of the `ImportDataStoreFile` message response changed at some point, but the importDataStore code wasn't updated.

So I did a slightly larger refactoring to get the types to line up.  And I dropped the onSuccess and onFailure callbacks, which were only doing logging, and moved the logging into the `Try` result handling.


jfalkner:
Nice work @skinner! Fixes #205 too and explains the error that was being seen as well.

PR #211, opened 2016-06-29
Dataset report fixes

natechols:
It's gross but it works.  Here's how I test it using PacBioTestData after calling `sbt smrt-server-analysis/run`:

```
./smrt-server-analysis/target/pack/bin/pbservice import-dataset --host localhost --port 8070 `pbdata get subreads-sequel`
./smrt-server-analysis/target/pack/bin/pbservice import-dataset --host localhost --port 8070 `pbdata get subreads-xml`
```

The first SubreadSet includes an sts.xml and has pbreports reports run, the second does not.  I can easily turn this into an automated test using the new pbtestkit-service-runner - not sure where this should live though.

### DIFF
--- a/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/reports/DataSetReports.scala
+++ b/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/reports/DataSetReports.scala
@@ -3,6 +3,8 @@ package com.pacbio.secondary.analysis.reports
 import java.nio.file.{Path, Paths}
 import java.util.UUID
 
+import collection.JavaConversions._
+
 import org.joda.time.{DateTime => JodaDateTime}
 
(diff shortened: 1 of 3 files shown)


smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/reports/DataSetReports.scala
```diff
@@ -69,9 +70,29 @@ object DataSetReports {
     val rptParent = jobPath.resolve(reportPrefix)
     rptParent.toFile().mkdir()
 
+    // all of the current reports will only work if at least one sts.xml file
+    // is present as an ExternalResource of a SubreadSet BAM file
+    val hasStatsXml: Boolean = dst match {
```
mpkocher: In a future iteration, you might want to consider using for comprehensions with Option + flatMap here.

PR #216, opened 2016-07-06
Remove blocking call in pbsmrtpipe job type

mpkocher:


mpkocher:
`stress.py` works for me locally.

PR #218, opened 2016-07-07
disable system job summary tick

skinner:
Before this change, I was seeing 12 to 50 of `Ask timed out` errors when I ran `pbtestkit-service-multirunner service-test-tiny-jobs.txt` with a copy of the smrtlink-beta database (on NFS).

With this change, I see two `Ask timed out` errors (I've run one complete test run so far).

The periodic GetSystemJobSummary tick would query for all the engine jobs and all the datastore files (two of the larger tables in our DB).  If a separate query was queued behind those queries, it could run into the ask timeout.

AFAIK most developers are working on an SSD and/or with a small database, and you wouldn't hit those timeouts in that case.

We should increase the timeouts as well; I'll make a separate PR for that.


jfalkner:
It seems unintuitive to me that these ticks are so expensive that queries behind it hit timeouts. Did you time the queries to estimate how long they take? In any case, reducing the 12 to 50 timeouts you were seeing seems likes a good win.

Is there any downside to disabling these ticks?

PR #224, opened 2016-07-12
Make insert mock data configurable.

mpkocher:


skinner:
> can you modify the defaults in the config that would be reasonable upper bound for testing on NFS?

Sure.  I think we'd want to replicate beta as closely as possible, so I'm inclined to just make the row counts be like the beta DB.

PR #226, opened 2016-07-13
Add SocketDiagnostics state to PBDM

smcclellan7:


mpkocher:
Please regenerate the java classes and commit them so there's a consistent checks from the XSD and java class at all times. Ideally only commit the changes of the java classes that don't have timestamp changes.

PR #228, opened 2016-07-13
modernize and clean up reference dataset generation code

natechols:
I left the code to convert ReferenceInfo to ReferenceSet in place for now, but the old reference uploader and associated cruft is gone.


smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/converters/FastaConverter.scala
```diff
@@ -1,138 +1,147 @@
...
-    f.close()
-
-    if (errors.isEmpty) {
-      Right(contigs)
-    } else {
-      Left(InValidFastaFileError(errors.toString()))
+  // FIXME workaround for lack of DataSetType.setDataSetMetaData
+  // this is a little hacky - there is probably a cleaner way to do it with
+  // implicits and/or type classes
+  protected def setMetadata(ds: T, metadata: U): Unit
```
natechols: I would love to know what the proper way to deal with this is.

smrt-analysis/src/test/scala/ConvertUtilsFastaValidateSpec.scala
```diff
@@ -20,8 +23,8 @@ class ConvertUtilsFastaValidateSpec extends Specification with LazyLogging{
     logger.debug(s"validating fasta file ${uri.toString}")
     val path = Paths.get(uri.toURI)
     validateFastaFile(path) match {
-      case Some(err) => false
-      case _ => true
+      case Left(err) => false
```
mpkocher: `validateFastaFile(path).isRight)`

natechols:
There's actually a bug running import-fasta through pbservice, so I'll have to fix that before we can merge.

PR #232, opened 2016-07-14
Add db summary tool

mpkocher:
There's still some friction over smrt-server-link and smrt-server-analysis DB URI.

For common  CLI interface, added explicit sys.exit call.


mpkocher:
As a second iteration, I think the import mock data interface should folded back into this tool as a subparser. Merging for now.

PR #236, opened 2016-07-15
Correctly update collections

smcclellan7:


mpkocher:
Please add a unittest

smcclellan7:
Yeah, I'm trying to figure out how to do that.  There's already a unittest that I thought covered this; I'm not sure why it was passing.

mpkocher:
Maybe just load Mark's XML and make sure you can parse it with your code? 

Those data models have everything as optional, I assume that's creating some friction.

smcclellan7:
Yep, I can confirm that smrtlink fails on Mark's XML without these changes, but succeeds with them.

PR #237, opened 2016-07-15
Tidy and fix warnings

mpkocher:


mpkocher:
@natechols Probably should take a closer look at this.

natechols:
LGTM - I won't second-guess the compiler.

PR #238, opened 2016-07-15
renumber V10 migration to V11.1

skinner:


skinner:
Just tried running smrtflow HEAD on a recent copy of the beta database, and found that V10 wasn't running.  This should fix that.

I don't believe this change would leave any running smrtlink instance database in a problematic state; the cases I know about are:
beta: hasn't run V10, so renumbering V10 doesn't affect it
alpha: hasn't run V10, so renumbering V10 doesn't affect it
anything running recent smrtflow: will already have run V12, so wouldn't run V11.1

PR #239, opened 2016-07-15
Add DatabaseSpec

smcclellan7:


smcclellan7:
@mpkocher does this look ok?

mpkocher:
Looks good to me, @skinner, can you take a look? I think you have more insight into the recent migration issues.

skinner:
I think this accomplishes the goal.  I'm a little worried about the maintenance burden here, where it looks like we'll have to change this spec for every migration.  I'm still hoping to find a way to do this generically, but that can be a separate PR.

Side issue: I also worry a little about all the timeouts.  If we were using blocking APIs, they'd all implicitly have infinite timeouts.  So I think it would be all right to use Duration.Inf more in our code, unless there's a specific goal that we want to achieve with finite timeouts.  But these look generous enough that I think they shouldn't cause intermittent test failures.

smcclellan7:
It doesn't seem like a bigger burden than any other unit test.  It's not like we add ten migrations a day or anything.  I actually kind of like the fact that you have to change it with every migration.  The fact that it's easy to break makes me more confident that it's actually testing what it purports to test.  A unit test that never needs updating and passes after every code change sounds suspicious to me.

As for the timeouts, I guess I put them in there because I didn't want the test to be able to hang forever, thereby preventing other tests from running.  But I'm happy to take 'em out, if you don't think we need them.

skinner:
> A unit test that never needs updating and passes after every code change sounds suspicious to me.

Well, I'm hoping to have a test that never needs updating and only breaks if you add a migration and forget to update TableModels.scala (or vice versa, or update them both in incompatible ways).  I think we also want a test that tests mainline code migrations against a database from the previous branch.  That test presumably could use this same code.  So there's a variety of useful ways for the test to fail; it's not necessarily always passing (which I agree would be less useful).

Agreed that changing the test on migrations isn't a big deal, though.

Agreed on not wanting the test to hang forever, I just worry about tests intermittently failing.  Maybe there's a way to balance those concerns?  I didn't mean for that last paragraph in my previous comment to be a hard objection--more of a thing I'd like us to consider.  Go ahead and leave the timeouts.

PR #240, opened 2016-07-17
Fix example tool

mpkocher:


mpkocher:
The "number" usage from the JSONSchema model still needs to be removed. 

It's currently (incorrectly) used by many of the TCs in pbsmrtpipe should be removed. This will require regenerating all the TCs and coordinating changes with the UI.

PR #242, opened 2016-07-19
Add db test on startup

smcclellan7:


smcclellan7:
@mpkocher how does this look?

mpkocher:
If that future fails does that stop the system from starting up? Or does this system still startup?

smcclellan7:
Hm, I'm fiddling around with it, and I can get the akka system to shutdown, but I can't figure out how to get the jvm to shutdown, if that's what you're looking for.  I could stick a System.exit call in there, but that feels ugly.  What do you think?

mpkocher:
I would suggest making a call the JobsActor on startup and shutdown if that future fails. What do you think?

mpkocher:
Can you test this model by starting up a fresh system, shutting it down, delete "uuid" from a table, then start the system up. It should fail to start up and return a non-zero exit code.

smcclellan7:
This is working for me, i.e. failing when the db is in an inconsistent state, passing otherwise. I'm not crazy about having to force the system to exit like that, but if I just throw an exception, Akka tries to keep everything running and the exception never reaches the top.

mpkocher:
https://github.com/PacificBiosciences/smrtflow/blob/master/smrt-server-base/src/main/scala/com/pacbio/common/app/BaseSmrtServerApp.scala#L159

There's a hook on `start`. Make a call to `jobsDaoActor ? GetSummary` after the successful bind, then kill things at that level.

mpkocher:
@smcclellan7 Can you update this?

smcclellan7:
I can move the test to the startup() method, but what's the purpose of going through the Akka layer?

mpkocher:
That works too. I was still think in terms of the old slick 2 model.

PR #243, opened 2016-07-19
Add SensorDiagnostics state to XSDs and Java

smcclellan7:


smcclellan7:
@mpkocher is this correct?

mpkocher:
This looks incorrect to me. 

- Remove any files that only have timestamp changes
- There's elements that have been removed for some reason
- the package-info.java needs special attention. These have custom namespaces set. You'll need to revert any changes to those files

mpkocher:
@smcclellan7 can you update this so we can merge this back?

PR #249, opened 2016-07-21
add Java code from IGV to create Fasta index

natechols:
removes samtools dependency.  This LGPL licensed, I left the original code header intact but we should probably mention this somewhere more central too.


natechols:
I'll try again after lunch.

PR #252, opened 2016-07-21
Add SLIA Client and CLI tool

mpkocher:
Still WIP, CLI submit doesn't work yet.

### DIFF
--- a/smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/SmrtLinkAnalysisInternalConfig.scala
+++ b/smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/SmrtLinkAnalysisInternalConfig.scala
@@ -1,0 +1,22 @@
+package com.pacbio.secondaryinternal
+
+import java.nio.file.{Path, Paths}
+
+import com.pacbio.common.dependency.Singleton
+import com.pacbio.secondary.analysis.configloaders.ConfigLoader
+
+/**
+  * Created by mkocher on 7/21/16.
+  */
+trait SmrtLinkAnalysisInternalConfig extends ConfigLoader{
+
+  lazy val reseqConditionsDir = Paths.get(conf.getString("smrt-server-analysis-internal.conditions-dir")).toAbsolutePath
+}
+
+// Trying to not infect the entire code base with Singletons
+// This entire model needs to go away.
+trait SmrtLinkAnalysisInternalConfigProvider extends ConfigLoader{
+
+  val reseqConditions: Singleton[Path] =
+    Singleton(() => Paths.get(conf.getString("smrt-server-analysis-internal.conditions-dir")).toAbsolutePath)
+}
--- a/smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/client/InternalAnalysisServiceClient.scala
+++ b/smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/client/InternalAnalysisServiceClient.scala
@@ -1,12 +1,14 @@
 package com.pacbio.secondaryinternal.client
 
+import java.io.IOError
 import java.net.URL
-import java.nio.file.Paths
+import java.nio.file.{Files, Path, Paths}
 import java.util.UUID
 
 import org.joda.time.{DateTime => JodaDateTime}
 
 import akka.actor.ActorSystem
+
 import com.pacbio.secondary.analysis.constants.FileTypes
 import com.pacbio.secondary.analysis.datasets.DataSetMetaTypes
 import com.pacbio.secondary.analysis.datasets.DataSetMetaTypes.DataSetMetaType
@@ -20,6 +22,9 @@ import com.typesafe.scalalogging.LazyLogging
 
 import scala.concurrent.Future
 
+import spray.http._
+import spray.client.pipelining._
+
 
 class InternalAnalysisServiceClient(baseUrl: URL)(implicit actorSystem: ActorSystem)
     extends AnalysisServiceAccessLayer(baseUrl)(actorSystem) with LazyLogging {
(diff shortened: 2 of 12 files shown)


jfalkner:
Some notes grouping together the file changes would be helpful, especially since this could be a good PR to reference when making more client API + CLI. For example, the general strategy appears to be `smrt-server-xxx` for the server components (stuff we already have) and `smrt-client-xxx` for the matching client. In addition to that the new client lib then needs a few other tweaks: Makefile, setup-tools-env.sh, build.sbt, etc.

smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/SmrtLinkAnalysisInternalConfig.scala
```diff
@@ -0,0 +1,22 @@
...
+  * Created by mkocher on 7/21/16.
+  */
+trait SmrtLinkAnalysisInternalConfig extends ConfigLoader{
+
+  lazy val reseqConditionsDir = Paths.get(conf.getString("smrt-server-analysis-internal.conditions-dir")).toAbsolutePath
+}
+
+// Trying to not infect the entire code base with Singletons
+// This entire model needs to go away.
+trait SmrtLinkAnalysisInternalConfigProvider extends ConfigLoader{
```
jfalkner: Any ideas for a work-around that'd let us convert over a few traits to be non-provider based? That'd let new stuff move away and enable PRs to del a few providers and not require a huge PR with all at once.

smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/client/InternalAnalysisServiceClient.scala
```diff
@@ -0,0 +1,111 @@
...
+        case _ => Future.failed(throw new Exception(s"Failed resolve Entry Point type $datasetMetaType"))
+      }
+    }
+
+
+    /**
+      * This needs to have a more robust implementation
+      *
+      * - Resolve Entry points looks for a SubreadSet and ReferenceSet
+      * - Does not valid resolved paths
```
jfalkner: nit: s/valid/validate/ ?

smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/tools/InternalAnalysisClientToolApp.scala
```diff
@@ -0,0 +1,190 @@
...
+    } text "Hostname of smrtlink server"
+
+    opt[Int]("port") action { (x, c) =>
+      c.copy(port = x)
+    } text "Services port on smrtlink server"
+
+
+    opt[Unit]("debug") action { (_, c) =>
+      c.asInstanceOf[LoggerConfig].configure(c.logbackFile, c.logFile, debug = true, c.logLevel).asInstanceOf[CustomConfig]
+    } text "Display debugging log output"
```
jfalkner: Why is `opt[Unit]("debug")` needed here? `LoggerOptions.add(this.asInstanceOf[OptionParser[LoggerConfig]])` will add all of the CLI logging related options, including `--debug`.

smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/tools/InternalAnalysisClientToolApp.scala
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    // Submit Reseq Conditions JSON
+    arg[File]("reseq-json") action { (x, c) => c.copy(pathToReseqConditions = x.toPath)} text "Path to Reseq Condition JSON"
+
+    opt[String]("name") action { (x, c) => c.copy(jobName = x)} text "Job Name"
+
+    opt[String]("host") action { (x, c) => c.copy(host = x)} text "Service Host"
+
+    opt[Int]("port") action { (x, c) => c.copy(port = x)} text "Service Port"
```
jfalkner: I'm running it to confirm, but it looks like there are duplicate `host` and `port` params. I'm not sure what happens if that is the case, I assume the later one overrides. 

Probably makes sense to have the common client code refactored somewhere. e.g. something like `SmrtClientOptions.add(...)` used similar to `LoggerOptions.add(...)` so that host, port and any shared auth or stuff like that can be in one spot and has consistent docs and usage.

smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/tools/InternalAnalysisClientToolApp.scala
jfalkner: Weird, it comes out like below when. I think you can delete it since it doesn't hook in a new method.

```
  --debug
        Display debugging log output
... // std other logger options
  --debug
        Same as --log-level DEBUG
```

smrt-server-analysis-internal/src/main/scala/com/pacbio/secondaryinternal/tools/InternalAnalysisClientToolApp.scala
natechols: I agree that it should be refactored, unfortunately this fits poorly with the scopt model - we'd have to make everything look like the logger options, so it's going to be relatively bloated.

jfalkner: The full output looks like below. [scopt docs](https://github.com/scopt/scopt) recommend nesting the params for the commands, but otherwise maybe the repeat is intended so that those options render under the command.
```
SMRT Link Client Internal Analysis Tool
    
  --host <value>
        Hostname of smrtlink server
  --port <value>
        Services port on smrtlink server
  --debug
        Display debugging log output
  --log2stdout
  --log-level <value>
  --debug
        Same as --log-level DEBUG
  --quiet
        Same as --log-level ERROR
  --verbose
        Same as --log-level INFO
  --log-file <value>
  --logback <value>
Command: status

  csv
        Path to Reseq Conditions CSV
  --reseq-json <value>
        Path to Output Reseq Conditions JSON
Command: convert

  reseq-json
        Path to Reseq Condition JSON
  --name <value>
        Job Name
  --host <value>
        Service Host
  --port <value>
        Service Port
Command: submit
```

jfalkner:
After reading it through and thinking about it some. I think it makes sense. 

Per client
- Make a `...Client` trait. e.g. `InternalAnalysisServiceClient`
- Make a `...ClientToolApp` object. e.g. InternalAnalysisClientToolApp
  - Include in the .scala a CLI parser, config, and something like `Modes` so that a Mode maps to the client API. e.g. `InternalAnalysisClientToolParser`, `CustomConfig` and `Modes.STATUS | CONVERT | SUBMIT`.

This strategy then gives us an easy to use Scala API or CLI. Seem pretty usable to me. The main design thing I wonder is if there is a clean way to have each client method encapsulate its respective name, config required (beyond host:port) and response serialization options. You could then break each out each as a trait or maybe even avoid needing dedicated `App` classes with all the Mode, config, etc. Have a reusable App that just has all the traits and knows how to populate opts for a Parser and uses `runAwait`, `runAwaitWithActorSystem`, and such.

In any case, a little logic that seems ripe for reuse. Any way to put this in a trait of otherwise recycle may help clean up the Client and App so it more succinctly shows the logic specific to that client.

- `runAwait`, `runAwaitWithActorSystem` and `runner` to a reusable trait?
- implicit or otherwise provide a shared, succinct way to convert (host, port) to a URL. `convertToURL` and `toUrl` to the same thing. Not a big deal, but it seems like every client will use this method.

All optional suggestions above. Overall, looks good to me.

jfalkner:
Thoughts on how to test these clients? Would be helpful to have examples of that too.

mpkocher:
This is growing organically from the original `pbservice` python port to scala client tool. See the `PbService.scala` for examples.

After looking at the port, I noticed a few common items and tried to put them in a central place. There's more opportunity here, I would like to get @natechols feedback as well.

PR #253, opened 2016-07-22
Endpoints reply with json instead of string

smcclellan7:


smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/services/DataSetService.scala
```diff
@@ -124,11 +122,9 @@ class DataSetService(dbActor: ActorRef) extends JobsBaseMicroService with SmrtLi
         } ~
         path(DETAILS_PREFIX) {
           get {
-            respondWithMediaType(MediaTypes.`application/json`) {
-              complete {
-                ok {
-                  (dbActor ? id.map(GetDetailsById, GetDetailsByUUID)).mapTo[String]
-                }
+            complete {
```
mpkocher: I believe this is returning the DataSet file after it's converted to JSON (as a string). Hence the explicit media type. This shouldn't return a MessageResponse.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/services/JobService.scala
```diff
@@ -124,11 +125,9 @@ trait JobService
     } ~
     path(IntNumber / JOB_REPORT_PREFIX / JavaUUID) { (jobId, reportUUID) =>
       get {
-        respondWithMediaType(MediaTypes.`application/json`) {
-          complete {
-            ok {
-              (dbActor ? GetDataStoreReportByUUID(reportUUID)).mapTo[String]
-            }
+        complete {
```
mpkocher: This is returning a PacBio JSON Report from the filesystem. This shouldn't return the MessageResponse instance.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
```diff
@@ -964,38 +964,35 @@ trait DataSetStore extends DataStoreComponent with LazyLogging {
       q.result.map(_.map(toDataStoreJobFile))
     }
 
-  def getSystemSummary(header: String = "System Summary"): Future[String] = {
-
+  def getSystemSummary(header: String = "System Summary"): Future[MessageResponse] = {
```
mpkocher: Please revert this. The spirit of this feature request was for all **Service** responses to return JSON.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/services/JobService.scala
```diff
@@ -151,11 +152,9 @@ trait JobService
     } ~
     path(IntNumber / JOB_OPTIONS) { id =>
       get {
-        respondWithMediaType(MediaTypes.`application/json`) {
-          complete {
-            ok {
-              (dbActor ? GetJobById(id)).mapTo[EngineJob].map(_.jsonSettings)
-            }
+        complete {
```
mpkocher: Does this need .parseJson? The raw value is a string.

mpkocher:
@skinner or @natechols Can you talk a look at this? There's a lot of endpoints that have been changed. 

The original implementation should have perhaps explicitly parse the json if the output was pulled from a file (instead of using strings to make the code easier to refactor).

PR #259, opened 2016-07-26
smrt-server-lims code. Ready for code review

jfalkner:
CC @mpkocher 

This is a rebase of all the smrt-sever-lims server-side code in its current state. Pair this with [jfalkner/SmrtLimsUI](https://github.com/jfalkner/SmrtLimsUI) for example client-side code in HTML/CSS/JS + ReactJS.

See the [README.md](https://github.com/PacificBiosciences/smrtflow/blob/77406002fbe657d87c6a77683ceeec2c8d42809a/smrt-server-lims/README.md) for how to run, test, example use cases and link to the drafted [specification.md](https://github.com/PacificBiosciences/smrtflow/blob/77406002fbe657d87c6a77683ceeec2c8d42809a/smrt-server-lims/README.md). The `./scripts` dir has helper scripts that use `find` and `curl` to scan for all PacBio `lims.yml` files and import them to the service -- the main use case tested.

The core web services are exercised by two tests (as noted in [README.md](https://github.com/PacificBiosciences/smrtflow/blob/77406002fbe657d87c6a77683ceeec2c8d42809a/smrt-server-lims/README.md#tests)). [RouteImportAndResolveSpec]() exercises the API in a linear fashion and is a good bit of code to read first.
## UI Example

Here is an example of the UI that is using this PR's code as the server-side backing. Very similar to the old UI that relied on ES. The UI is easy to update but not part of an official build.

screen shot 2016-07-25 at 1 34 31 pm

In order to replicate the above, host the static web UI straight from the repo.

``` bash
# Host this checkout as a website
git clone https://github.com/jfalkner/SmrtLimsUI.git
cd SmrtLimsUI
python -m SimpleHTTPServer
```

In a different terminal, checkout and run the server-side code.

``` bash
git clone <this_repo>
sbt clean compile smrt-server-lims/run
# could also use `sbt smrt-server-lims/assembly` then run the JAR 
```

Now the server should be available on `http://127.0.0.1:8070`. Use curl to copy over data from `/pbi/collections`.

```
# use curl via helper script. see script for details
cd smrt-server-lims/scripts
./import_lims_.sh /pbi/collections/312/3120145/r54009_20160426_164705/1_A01/lims.yml
```

Now visit the same URL as in the screenshot: `http://localhost:63342/SmrtLimsUI/index.html?q=3120145`


jfalkner:
@mpkocher I'm out of time to add the few smaller updates we talked about, but I'll list them here.

- Add fixes for BZ tickets 33829, 33843 and 33844 (aka rest of my queue). The UI has all the core bits working as noted in the description.
- Few small updates to `specification.md`
  - Add what is pulled from `lims.yml`, namely this logic.
  - Add how the subreadset is added via directory lookup in this code. This needs an update to explicitly first lookup based on the context name.
- Client-side code modeled after the other internal example last week. Should be no prb, but @mpkocher to write up a brief spec with examples of CLI use. This may entirely replace the `./scripts` dir I have here.

mpkocher:
The CLI docs are very minimal, therefore I'm just going to add them inline.

Import either a single lims.yml file or the code will recursively walk the file system if a directory is provided.
```
$> smrt-client-lims import /path/to/lims.yaml --host my-host --port 8081
```
Or
```
$> smrt-client-lims import /pbi/collections/1234 --host my-host --port 8081
```

Get SubreadSet by RunCode. Returns a summary of the LimsSubreadSet or return a non-zero exit code.

```
$> smrt-client-lims get-runcode <RUN_CODE> 
```

Resolver

Returns a DataSetMetadata summary

```
$> smrt-client-lims resolve my-lambda-neb
```

Version 2 could support getting a SubreadSet by UUID, or searching by experiment id. 

See the SLIA CLI tool for an example.

mpkocher:
LimsSubreadSet 

- runcode (lims.yml)
- experiment id (lims.yml)
- SubreadServiceDataSet (see model in smrtflow, this will give it parity with the SMRT Link and SMRT Link Analysis DataSet services)
- pa version (from subreadset XML)
- ics version from subreadset XML)

smrt-server-lims/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+# run the LIMS services independent of the greater codebase
+sbt clean compile smrt-server-lims/run
+```
+
+You'll now have the service bound to port `8070`. The host, port and 
+location of the database can all be configured in `application.conf`.
+
+```
+smrt-server-lims {
```
mpkocher: Please use the pbservices namespace and override the db-uri value. This will enable the system to be configured similarly to SMRT Link Analysis and the Internal Analysis server.

smrt-server-lims/src/main/resources/application.conf
```diff
@@ -0,0 +1,10 @@
+akka {
+  loglevel = ERROR
+  event-handlers = ["akka.event.slf4j.Slf4jEventHandler"]
+}
+
+smrt-server-lims {
```
mpkocher: Please use the pbservices namespace and override the db-uri value. This will enable the system to be configured similarly to SMRT Link Analysis and the Internal Analysis server.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/InternalServiceActor.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+class InternalServiceActor extends Actor
+    with ConfigLoader
+    with JdbcDatabase
+    with DefaultDatabase
+    with FileLookupSubreadset
+    with ImportLims
+    with ResolveDataSet {
+
+  lazy val jdbcUrl: String = conf.getString("smrt-server-lims.jdbc-url") // required for JdbcDatabaseService
```
mpkocher: It would be useful to consider explicitly initializing the Actor the db-uri. This would enable easier testing.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/Models.scala
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Superset of a SubreadDataSet file that represents lims.yml
+ *
+ * This is a practical superset of the subreaddataset XML file data. It includes various information that gets
+ * calculated and included in the lims.yml files MJ makes. Long-term, this abstraction needs to be rethought and likely
+ * recast to a more formal abstraction. As-is, this data duplicates other
+ */
+case class LimsSubreadSet(
+    val uuid: String,
```
mpkocher: In general, I would encourage using types if possible. In this case, using UUID for `uuid`. 

The motivation for the use of "json" here in this case class is unclear.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/database/Database.scala
```diff
@@ -0,0 +1,39 @@
...
+import spray.json.JsValue
+
+
+/**
+ * All of the public DAO methods
+ *
+ * This trait is the main DB-agnostic interface that the rest of the codebase relies on. The Cake
+ * pattern is used to build up the desired DB implementation at runtime.
+ */
+trait Database {
```
mpkocher: UUIDs over String would be preferred.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/database/h2/H2Database.scala
```diff
@@ -0,0 +1,174 @@
...
+ * `H2CreateTables.sql` will create needed tables and extra indexes for good performance. It is safe
+ * to re-run since all statements use `IF NOT EXISTS`.
+ *
+ * All of the xxxT named String instances are SQL templates for the `PreparedStatement` instances, and
+ * they are also used as a simply way to synchronized access. `aliasSelectT` is the only template
+ * with a sub-query. It is done this way to efficiently lookup aliases and avoid multiple calls to
+ * the DB.
+ */
+trait H2Database extends Database {
+  this: JdbcDatabase =>
```
mpkocher: Is there a plan in the next iteration to migrate this to Slick?

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ResolveDataSet.scala
```diff
@@ -0,0 +1,96 @@
...
+                  ctx.complete(if (lys.nonEmpty) 200 else 404, lys.toJson.prettyPrint)
+              }
+            }
+        }
+      }
+    }
+  }
+
+  // TODO: move this to an actor, probably too slow to keep on request dispatcher
+  def resolveLimsSubreadSet(q: String): Seq[LimsSubreadSet] = {
```
mpkocher: Leveraging the spray matchers would be useful here. Specifically, the Int, regex and UUID matchers. This would remove the use of the nested Try.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ResolveDataSet.scala
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * GET /smrt-lims/resolver/{dataset-type-short-name}/{name-id}
+   * POST /smrt-lims/resolver/{dataset-type-short-name}/{UUID} name="name-id"
+   * DELETE /smrt-lims/resolver/{datasetyp-type-short-name}/{name-id} # "Unregister `name-id` to specific SubreadSet"
+   */
+  val resolveDataSetRoutes =
+    path("resolver" / Segment / Segment ) { // TODO: can specify dataset-type-short-name as regex
+      (dt, q) => {
+        // GET = resolve the alias to the dataset type
+        get {
```
mpkocher: This looks very SubreadSet specific. The intended model is for any dataset type to have an alias. For example, "lambaNeb" to resolve the ReferenceSet.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ResolveDataSet.scala
```diff
@@ -0,0 +1,96 @@
...
+          complete(subreadByAlias(q).toJson.prettyPrint) // TODO: match `dt` and do dataset-type-short-name specific alias
+        } ~
+        // PUT = set the alias for the dataset type
+        post {
+          parameters('name) { name =>
+            Try(setAlias(name, q, "lims_subreadset")) match {
```
mpkocher: This could benefit from leveraging map/flatMap.

```
Welcome to the Ammonite Repl 0.6.2
(Scala 2.11.8 Java 1.8.0_71)
mkocher-mkocher@ import scala.util.Try
import scala.util.Try
mkocher-mkocher@ Try { "x" }.map(_ => "completed").getOrElse("completed(500, 'bad'")
res1: String = "completed"
```

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/MainSimple.scala
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Entry point for the server that is as stripped-down as possible. The idea was to build up from
+ * the minimum and see what is needed.
+ */
+object MainSimple extends App with ConfigLoader with LazyLogging{
+
+  LoggerOptions.parseAddDebug(args)
+  logger.debug(s"Starting smrt-lims with arguments: ${args.mkString(" ")}")
+
+  implicit val system = ActorSystem("internal-smrt-link-system")
```
mpkocher: `ActorSystem("lims")`

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/InternalServiceActor.scala
jfalkner: I'm not following. Can you explain this suggestion better. With an example snippet, if it makes sense?

As is, the code uses the `Database` trait to abstract the API then the `H2Database` + `JdbcDatabase` traits to satisfy the implementation. The `jdbcUrl` is declared by `JdbcDatabase` and overriding it to be the in-memory H2 instance is what allows easy testing. Example in `TestDatabase` in the test code package.

Long way to say that as-is it has an easy way to do in-mem test db use, production DB config, and even file-backed db use for the stress test. README.md has some notes on these uses.

smrt-server-lims/src/main/resources/application.conf
jfalkner: Fixed as suggested in aa72832

Fixes BZ [#33843](https://bugzilla.nanofluidics.com/show_bug.cgi?id=33843)

For consideration. I don't recall JDBC having a concept of a URI. JDBC URLs are used to configure drivers; however, the `pb-services` using the confusing `db-uri` name. Everything works fine as-is, but it might be helpful to change the nomenclature if `db-uri` always means JDBC URL.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/Models.scala
jfalkner: I understand. It is a quick change to enforce this and I'll roll it in. I don't follow what we gain by treating this as a formal UUID. It will restrict what imports are allowed. Right now there are many cases in `/pbi/collections` where no `.subreadset.xml` exists. I'm not aware of any cases where the file exists but it isn't a UUID, but that too could cause the same problem.

The problem is that without a UUID then the the pk for the DB table fails. As-is the `String` use is in the implementation and allows for flexibility to either import non-UUID values (if any use cases merit this) and for datasets without `.subreadset.xml` files to be imported with `runcode` acting as the temp UUID. The pk could always be switched too.

Performance wise, UUID and String should be equivalent in an RDMS. It is just syntax we're enforcing.

jfalkner: Swap to UUID added as suggested in 8f667b6

mpkocher: This really shouldn't be motivated by the database layer. 

This model is independent should be using proper types to clearly communicate the interface. For example, if it's a data time, then it should be using DataTime. If it's UUID is required by the model, then enforcing this at the type level is preferred. 

If there isn't a subreadset.xml, then the import should fail.

jfalkner: > The motivation for the use of "json" here in this case class is unclear.

The data used for indexing and fast lookups is explicitly modeled. The rest is less formally modeled because it is unclear how it needs to be used and stashing it as a map allows for flexibility. Adding a `version` field (right now it is v0) allows for enforcing required values in the JSON, registering a validation class and providing data migration. I figured we could use this approach unless it ends up that formalizing a few fields is all that is really need.

One thing I haven't been able to do is compare all the `lims.yml` files and `.subreadset.xml` files to see if there are different versions and if we care. I'm not yet sure where it is appropriate to draw the lines regarding what data is supported by the resolution service and what is not.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ImportLims.scala
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Trait required for abstracting the UUID file lookup for prod vs testing
+ */
+trait LookupSubreadset {
+  def subreadset(path: String): Option[SubreadSet]
+}
+
+trait FileLookupSubreadset {
+  def subreadset(path: String): Option[SubreadSet] = {
```
mpkocher: Please use types. URL or Path explicitly here.

Also, `Try {}.toOption` would remove the case match.

smrt-server-lims/src/main/resources/application.conf
mpkocher: I wasn't referring to the JDBC URL, I was referring to the `FileLookupSubreadset` usage. I've added a comment below.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ImportLims.scala
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    // need the path in order to parse the UUID from .subreadset.xml
+    m.get("path") match {
+      case Some(p) => loadData(subreadset(p), m)
+      case None => "No path in lims.yml file. Can't attempt UUID lookup"
+    }
+  }
+
+  def loadData(s: Option[SubreadSet], ly: mutable.HashMap[String, String]) : String = {
```
mpkocher: Immutable types would be preferred here.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ImportLims.scala
```diff
@@ -0,0 +1,118 @@
...
+           "instid" -> c.getInstrumentId
+         ).toJson)
+      }
+      case None => (null, "{}".toJson)
+    }
+    val expid = ly("expcode").toInt
+    val runcode = ly("runcode")
+
+    // TODO: should be smarter about error handling here and not making alias if setSubread() fails
+    Try(if (uuid != null) setAlias(makeShortcode(uuid), uuid, LimsTypes.limsSubreadSet))
```
mpkocher: In general, please remove these uses of null in favor of Option.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ResolveDataSet.scala
mpkocher: Here's an example for the Regex matcher

```
        path(PREFIX / "runcode" / "^[0-9]{7}-[0-9]{4}$".r) { runCode =>
          get {
            complete {
              dao.getByRunCode(runCode)
            }
          }
        }
```

The IntNumber and JavaUUID are the other matchers.

http://spray.io/documentation/1.2.3/spray-routing/path-directives/pathmatcher-dsl/

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/database/Database.scala
jfalkner: Full UUID use as suggested in 708dab8

This also exposed a bug that was missed by the test suite where pre-UUID/`.subreadset.xml` parsing the runcode was defaulted as the UUID. Not sure how I missed it, but it is now fixed too.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/database/h2/H2Database.scala
jfalkner: I don't currently intend to, but will if it is a requirement. The impl works well as-is and I'm not sure it merits using Slick just because. 

After looking closely at lower level JDBC use vs Slick, I think a connection pool w/cached `PreparedStatement` objects is a clear winner, assuming you have a DAO abstracting away any RDMS logic. This impl doesn't use a pool but is effectively the same strategy. Slick is clever but has performance, transaction and object->schema gotchas that I'm not a fan of. It isn't used here, but finer grain control of transactions is also a perk that it seems Slick's abstractions complicate.

In any case, as long as indexes exist for all the searchable fields, this service's perf should be good enough using Slick or this JDBC route.

jfalkner:
Done as suggested in 7f0e0fc.

The main reason I demux'd in `resolveLimsSubreadSet` was so that a single URI could auto-resolve any ID, including any fuzzy matching style logic similar to the prior ElasticSearch impl/demo. That doesn't seem to be a needed use case though. The refactor has explicit URIs per field you can lookup and should perform better since it doesn't attempt up to 3 queries.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ResolveDataSet.scala
jfalkner: Done as suggested in 7f0e0fc.

The main reason I demux'd in `resolveLimsSubreadSet` was so that a single URI could auto-resolve any ID, including any fuzzy matching style logic similar to the prior ElasticSearch impl/demo. That doesn't seem to be a needed use case though. The refactor has explicit URIs per field you can lookup and should perform better since it doesn't attempt up to 3 queries.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ResolveDataSet.scala
jfalkner: The URI is generic so that the middle segment can be any dataset. The second segement is the alias that maps to a UUID for that dataset.

I hard-coded a single `get` to handle subreadsets because that is the initial use case. I think the code can easily be extended to support the other dataset types.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ResolveDataSet.scala
jfalkner: I tried the suggestion. It resulted in the below code. Why is this a benefit? Is there more than just style and cramming it all on one line? If the text was shorter, then one line may be more readable. As-is the `match` seems more readable to me.

```scala
Try{ setAlias(name, UUID.fromString(q), "lims_subreadset" }.map(_ => complete("Set $dt as alias for $id")).getOrElse(complete(500, "Couldn't set alias '$id' for $dt"))
```

vs existing code

```
Try(setAlias(name, UUID.fromString(q), "lims_subreadset")) match {
  case Success(_) => complete("Set $dt as alias for $id") 
  case Failure(t) => complete(500, "Couldn't set alias '$id' for $dt")
}
```

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/MainSimple.scala
jfalkner: Done as suggested in 2d19f4c

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ImportLims.scala
jfalkner: Done as suggested in 3a91ba9

It'd be helpful if in the "Please use" style comments you note why, in particular if there is any benefit or drawback to the code other than a preference for stronger typing. Otherwise I'm not sure if there is anything to learn from the suggestion or if it is style preference enforcement.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ImportLims.scala
jfalkner: Switched to immutable in 48ea7e5

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ImportLims.scala
jfalkner: Done as suggested in 0fc034b

smrt-server-lims/src/main/resources/application.conf
jfalkner: I cleaned up that `FileLookupSubreadset` in the other comment. Thanks for clarifying.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/Models.scala
jfalkner: > This really shouldn't be motivated by the database layer.

Agreed, I swapped the index in the previous commit so it should no longer be tied to it.

> If there isn't a subreadset.xml, then the import should fail.

Nice. I should have asked for clarity on that assumption earlier. That simplifies the code and data model.

jfalkner:
A full pass through the comments is done. Thanks for the review @mpkocher 

I still have to update the docs and add the CLI.

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/services/ResolveDataSet.scala
mpkocher: It doesn't have to be on one line. The common formatting model is:

```
val x = aList.map(_.name)
                  .find(_ == "Ralph")
                  .map(_.toUpperCase)
```
The benefit is in the clarity in chaining of operations by leveraging flatMap/map and friends.

jfalkner:
Modeling the CLI work after https://github.com/PacificBiosciences/smrtflow/pull/252

jfalkner:
@mpkocher Help please on two things. I put up the WIP code for adding the client in aafc69e

I was able to successfully run `import` via the client before hitting this. Import returns a `String`. I hacked in some tests in `LimsClientToolsAppTest`

Does the error below make sense to you? I'm not very familiar with how to set what is needed for the unmarshalling of a custom case class. The same class does get unmarshalled via spray-json in `ResolveDataSet`

screen shot 2016-07-28 at 12 35 47 pm

While implementing, I also realized that I'm unsure where the data comes from for this CLI command.

> Resolver

>Returns a DataSetMetadata summary

>```
$> smrt-client-lims resolve my-lambda-neb
```

**Comment by jfalkner** [comment]
83ec087 has the solution to the unmarshalling issues. Needed to import `import spray.httpx.SprayJsonSupport._`

The `resolve` CLI command noted above is still one I need some clarity on.

**Comment by jfalkner** [comment]
Some notes from the team meeting. The last feature noted above with the `DataSetMetadata` summary is being punted from this PR. We'll have a followup bug.

I'll do a final scan tomorrow, but the only thing left should be to finish up the CLI test code. I have been using a small hack to run it and it really should be in test specs.

**Comment by mpkocher** [comment]
It looks like the tests are not passing due to a SQL error.

```
[info] ! Import and be resolvable in a minimal stress test
[error]    JdbcSQLException: : Table "ALIAS" not found; SQL statement:
[error] MERGE INTO ALIAS (alias, uuid, type) VALUES (?, ?, ?) [42102-192]  (DbException.java:345)
[error] org.h2.message.DbException.getJdbcSQLException(DbException.java:345)
```

**Comment by jfalkner** [comment]
I think it is the same table create timing bug. I started a new test and am cleaning it up now.

**Comment by jfalkner** [comment]
> I'll do a final scan tomorrow, but the only thing left should be to finish up the CLI test code. I have been using a small hack to run it and it really should be in test specs.

OK, CLI tests are up. The tests confirm that both the API and CLI inputs and outputs work as expected. Should be solid code coverage of all the code.

**Comment by jfalkner** [comment]
Still working on making CircleCI pass. Looks like it is a `beforeAll` style fix again.

Code coverage is looking good. `smrt-server-lims` is already past 80%.

screen shot 2016-08-01 at 1 00 09 pm

Can replicate  by running.

```bash
# run all the tests for the sub-project -- notice build.sbt for the projects has "coverageEnabled := true"
sbt clean smrt-server-lims/test
# make the coverage XML and HTML reports
sbt smrt-server-lims/coverageReport
```

jfalkner:
Code coverage was handy. It showed that alias DELETE was broken and not correctly tests. I fixed it. It also shows that these cases aren't covered.

- [ ] Find `.subreadset.xml` file where the name doesn't match the movie context. Edge case from spec.
- [ ] Recursive scan of directory to find `lims.yml` then upload them.
- [ ] Related to the above but masked by other tests. Returning multiple `LimsSubreadSets`. e.g. multiple results with the same `expid`.
- (Ignore) JVM correctly invokes System.exit() after running the client
- (Ignoring) error handling for cases where 

I don't think it is critical to address these before merge; however, is probably easiest to add now vs later.

jfalkner:
>- [x] Find `.subreadset.xml` file where the name doesn't match the movie context. Edge case from spec.
- [x] Recursive scan of directory to find `lims.yml` then upload them.

Coverage is excellent. Multi-upload/download logic tested. Only thing left here is to figure out why CircleCI is failing on the CLI tests.

screen shot 2016-08-02 at 4 45 04 pm

jfalkner:
@mpkocher leaving a note that I worked through adding CLI testing w/buffer of stdout. It means both the the client API and CLI are fully testable. May be a helpful example for other CLI modeled after your example.

mpkocher:
I'll try to run the tests tomorrow. Currently, I'm seeing several errors where the services are 404'ing due to db related issues.

jfalkner:
I just merged the latest master and tests still work locally for me.

```
sbt smrt-server-lims/test

...
[info] Passed: Total 29, Failed 0, Errors 0, Passed 29
[success] Total time: 310 s, completed Aug 8, 2016 11:28:30 AM
```

smrt-server-lims/src/main/scala/com/pacbio/secondary/lims/InternalServiceActor.scala
```diff
@@ -0,0 +1,40 @@
...
+    with ConfigLoader
+    with JdbcDatabase
+    with DefaultDatabase
+    with FileLookupSubreadset
+    with ImportLims
+    with ResolveDataSet {
+
+  lazy val jdbcUrl: String = conf.getString("pb-services.db-uri") // required for JdbcDatabaseService
+
+  createTables
```
jfalkner: It'd be helpful to figure out how to get a better error message. Any ideas? Maybe this is it, but it is unclear what the root cause is.

mpkocher:
I'll try running the tests locally.

mpkocher:
I'm seeing errors, but the failures look different than the failures in Circle.

```
[info] x API should fail pre-import: lookup via expcode
[error]    '1' is not equal to '0' (CommandLineToolsSpec.scala:76)
[info] 
[info] x CLI should fail pre-import: get-expcode
[error]    ''
[error]    
[error]     is not equal to 
[error]    
[error]    'Failed to run spray.httpx.UnsuccessfulResponseException: Status: 404 Not Found
[error]    Body: []
[error]    ' (CommandLineToolsSpec.scala:78)
[error] Expected: Failed to run spray.httpx.UnsuccessfulResponseException: Status: 404 Not Found
[error] Body: []
[error] Actual:   
[info] 
[info] + API should fail pre-import: lookup via runcode
[info] + CLI should fail pre-import: get-runcode
[info] + Import data from POST
[info] + API lookup via expcode works
[info] x CLI get-expcode works
[error]    'List(LimsSubreadSet(8fed57d4-5233-4836-8c86-c005a2a5b711,3220001,3220001-0001,/tmp/mutli_import_and_results6803353945513953758/3220001-0001,3.0.17,3.0.5.175014,A01,m54009_160426_165001,2016-04-26T18:46:06Z,Inst54009,54009), LimsSubreadSet(5fe01e82-c694-4575-9173-c23c458dd0e1,3220001,3220001-0006,/tmp/CLI6380546418444418383,3.0.17,3.0.5.175014,A01,m54009_160426_165001,2016-04-26T18:46:06Z,Inst54009,54009))
[error]    '
[error]    
[error]     is not equal to 
[error]    
[error]    'List(LimsSubreadSet(5fe01e82-c694-4575-9173-c23c458dd0e1,3220001,3220001-0006,/tmp/CLI6380546418444418383,3.0.17,3.0.5.175014,A01,m54009_160426_165001,2016-04-26T18:46:06Z,Inst54009,54009))
[error]    ' (CommandLineToolsSpec.scala:92)
[error] Expected: ...dSet([5]fe[01e82]-[c694]-4[575]-[9173]-c[23c458dd0e]1,...00[6],/...p/[CLI63]80[]5[]4[64]1[]8[4444]1[8383],3...:4[6:06Z,Inst54009,54009))]
[error] Actual:   ...dSet([8]fe[d57d4]-[5233]-4[836]-[8c86]-c[005a2a5b71]1,...00[1],/...p/[mutli_import_and_results6]80[33]5[39]4[55]1[395375]8[/322000]1[-0001],3...:4[]
[info] 
[info] + API lookup via runcode works
[info] + CLI get-expcode works
[info] + API lookup via UUID works
[info] + CLI get-uuid works
[info] 
[info] Total for specification CommandLineToolsSpec
[info] Finished in 327 ms
[info] 11 examples, 3 failures, 0 error
[error] Failed: Total 29, Failed 3, Errors 0, Passed 26
[error] Failed tests:
[error] 	com.pacbio.secondary.lims.CommandLineToolsSpec
[error] (smrt-server-lims/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 335 s, completed Aug 8, 2016 8:57:18 AM
```

jfalkner:
Interesting. Not clear to me what is going wrong here. Ugh.

jfalkner:
db9805c fixes one bug in a test where the ordering of multiple uploads was incorrectly assumed. The ordering consistently stays the same on my laptop, but varies if run on others or login nodes. The test now allows for this.

smrt-server-lims/src/test/scala/com/pacbio/secondary/lims/RouteImportAndResloveSpec.scala
```diff
@@ -0,0 +1,151 @@
...
+      runcode mustEqual subreadsByRunCode(runcode).head.runcode
+    }
+    "runcode resolvable via GET" in {
+      Get(s"/smrt-lims/lims-subreadset/runcode/$runcode") ~> sealRoute(resolveRoutes) ~> check {
+        response.status.isSuccess mustEqual true
+        runcode mustEqual response.entity.data.asString.parseJson.convertTo[Seq[LimsSubreadSet]].head.runcode
+      }
+    }
+    "UUID resolvable via API" in {
+      uuid mustEqual (subread(uuid) match { case Some(ls) => ls.uuid })
```
mpkocher: Might want to consider adding a Some matcher to remove this compiler warning.

```
val x = subread.map(_.uuid) // Option[UUID]
Some(uuid) mustEqual x // can't remember the specific Some matcher
```

jfalkner: Updated as suggested in 44de7ef

PR #262, opened 2016-07-28
Fixed --log-file flag

jfalkner:
CC @skinner @mpkocher 

Fixed use of the `--log-file` flag. What happened is that a helper method that would auto-set `--debug` if no file or debug was set got out of sync with a change that did `s/--logfile/--log-file`.

Probably should have been covered by a test. Can file a ticket for that if desired later. This should fix it.

```
$ sbt "smrt-server-analysis/run --log-file /tmp/foo.log"
...
cat /tmp/foo.log

# notice, sbt uses the relative path as the sub-project
$ sbt "smrt-server-analysis/run --log-file foo.log"
...
cat smrt-server-analysis/foo.log
```

### DIFF
--- a/smrt-server-logging/src/main/scala/com/pacbio/logging/LoggerOptions.scala
+++ b/smrt-server-logging/src/main/scala/com/pacbio/logging/LoggerOptions.scala
@@ -66,7 +66,7 @@ object LoggerOptions {
   }
 
   def parseAddDebug(args: Seq[String]): Unit = {
-    val requireOne = Set("--logfile", "--log2stdout", "-h")
+    val requireOne = Set("--log-file", "--log2stdout", "-h")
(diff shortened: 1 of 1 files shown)


jfalkner:
I tested this by hand. I'll run the tests locally and if they are good, merge this when CircleCI is green.

The change for this PR is tiny. A string value that was already incorrect. So nothing should be able to break more.

jfalkner:
@skinner interestingly, `sbt "smrt-server-analysis/run fooe.log"` will make `smrt-server-analysis/foo.log`. It ends up that the working directory is that of the sub-project.

skinner:
> `sbt "smrt-server-analysis/run foo.log"` will make `smrt-server-analysis/foo.log`

Ah, so I guess I was using the wrong test to try to demonstrate the bug.  The real place I noticed it was during the stress testing; the makefile stress test rule passes an absolute path to `--log-file`.

PR #263, opened 2016-07-28
Add basic Simulator to smrtflow

smcclellan7:


jfalkner:
What is this? I just resolved a merge conflict on a branch and realized I had no idea what `smrt-server-sim` is for and there don't appear to be docs.

PR #264, opened 2016-07-29
Project api changes

skinner:
The original API had separate endpoints for changing the lists of users and datasets in a project.  But dbarreto objected on the grounds that the project page had one save button, and if multiple requests were required to save a project, it was awkward on the client to track the success/failure of partial project saves.  So this change makes it so that saving a project is a single request with a body that contains the project member/dataset lists.


skinner:
Tests pass, so I went ahead and merged this because there's some time pressure and because @mpkocher is on PTO today IIRC.

PR #269, opened 2016-08-01
Add simple Run Design scenario

smcclellan7:


smrt-server-sim/src/main/scala/com/pacbio/simulator/steps/HttpSteps.scala
```diff
@@ -0,0 +1,71 @@
...
+import scalaj.http.{Http, HttpOptions}
+
+trait HttpSteps {
+  this: Scenario with VarSteps =>
+
+  val defaultHost: String = null
+  val defaultPort: Int = 8080
+  val defaultTimeout: Int = 10000
+
+  object HttpStep {
```
mpkocher: Can you push this back to the common client layer? I'm not sure we want to use both spray.http and scalaj.

smrt-server-sim/src/main/scala/com/pacbio/simulator/scenarios/RunDesignScenario.scala
```diff
@@ -0,0 +1,69 @@
...
+ * Example config:
+ *
+ * {{{
+ *   smrt-link-host = "smrtlink-bihourly"
+ *   smrt-link-port = 8081
+ *   run-xml-path = "/path/to/testdata/runDataModel.xml"
+ * }}}
+ */
+object RunDesignScenarioLoader extends ScenarioLoader {
+  override def load(config: Option[Config])(implicit system: ActorSystem): Scenario = {
```
mpkocher: Can you comment on the motivation for Option[Config] instead of Config?

smrt-server-sim/src/main/scala/com/pacbio/simulator/steps/HttpSteps.scala
smcclellan7: Done.

smrt-server-sim/src/main/scala/com/pacbio/simulator/scenarios/RunDesignScenario.scala
smcclellan7: In general, a config may not be needed.  The ExampleScenario, for instance, does not require any configuration.

smrt-server-sim/smrt-server-sim.iml
```diff
@@ -0,0 +1,173 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mpkocher: Looks good, but can you remove this file?

mpkocher:
As a second pass, we should try to get CLI support for Run Design.

natechols:
Can we go ahead and merge?  I think the tests I'm supposed to add will build on this.

mpkocher:
I would like that iml to be removed before merging.

PR #272, opened 2016-08-04
Fix convert-rs-movie endpoint and add pbservice support

natechols:


smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
```diff
@@ -623,6 +642,54 @@ class PbService (val sal: AnalysisServiceAccessLayer,
...
+          case Success(jobInfo: EngineJob) => waitForJob(jobInfo.uuid)
+          case Failure(err) => errorExit(s"RS II movie import failed: ${err}")
+        }
+      case _ => errorExit(s"The path does not appear to be valid metadata XML")
+    }
+  }
+
+  private def listMovieMetadataFiles(f: File): Array[File] = {
+    (f.listFiles.filter((fn) =>
+      Try {
```
mpkocher: `Try {}.isSuccess`

PR #273, opened 2016-08-04
Add SL versions

mpkocher:


mpkocher:
Still need another pass at this, but I want to unblock the UI work.

PR #275, opened 2016-08-05
custom JSON handling for EngineJob model

natechols:
I double-checked - 'pbservice' works with the current server code and with smrtlink-nightly (which is still running 3.1).


natechols:
It turns out this is unnecessary, since the new fields are optional:

http://stackoverflow.com/a/10821816

PR #282, opened 2016-08-09
dataset import tests

natechols:
@mpkocher and @smcclellan7 you should probably review this

### DIFF
--- a/extras/run_sim_local.py
+++ b/extras/run_sim_local.py
@@ -14,8 +14,7 @@ import os
 import sys
 
 ROOT_DIR = op.dirname(op.dirname(op.abspath(__file__)))
-JAR_FILE = ROOT_DIR + \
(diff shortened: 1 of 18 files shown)


extras/run_sim_local.py
```diff
@@ -0,0 +1,85 @@
...
+import argparse
+import tempfile
+import shutil
+import os.path as op
+import os
+import sys
+
+ROOT_DIR = op.dirname(op.dirname(op.abspath(__file__)))
+JAR_FILE = ROOT_DIR + \
+    "/smrt-server-analysis/target/scala-2.11/smrt-server-analysis-assembly-0.1.5-SNAPSHOT.jar"
```
mpkocher: Can this hardcoded version be replaced in favor of a listdir or similar?

PR #283, opened 2016-08-09
Add IdAble model to start to replace Either[Int, UUID]

mpkocher:
Added example as a test.

### DIFF
--- a/smrt-common-models/src/main/scala/com/pacbio/common/models/Models.scala
+++ b/smrt-common-models/src/main/scala/com/pacbio/common/models/Models.scala
@@ -1,0 +1,32 @@
+package com.pacbio.common.models
+
+import java.util.UUID
+
+import scala.language.implicitConversions
+
+/**
+  * Created by mkocher on 8/9/16.
+  */
+object Models {
+
+  // "Old" model
+  case class MixedIdType(id: Either[Int, UUID]) {
+    def map[T](fInt: Int => _ <: T, fUUID: UUID => _ <: T): T = id match {
+      case Left(i) => fInt(i)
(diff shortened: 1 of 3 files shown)


natechols:
Will this automatically handle conversion to strings, e.g. when forming URLs in the client layer?

mpkocher:
It's still boxed if you call toString.

```
import Models._; import ModelImplicits._
val i: IdAble = 1
val u: IdAble = UUID.randomUUID()
u.toString
i.toString
history
@ u.toString 
res11: String = "UUIDIdAble(1770b34b-4624-405f-85c1-04198b078586)"
@ i.toString 
res12: String = "IntIdAble(1)"
```

I think overriding toString might be a bit too heavy handed. 

A util func can do the conversion. 

```
@ def idableToString(i: IdAble): String = i match {case IntIdAble(x) => x.toString; case UUIDIdAble(u) => u.toString} 
defined function idableToString
@ idableToString(i) 
res18: String = "1"
@ idableToString(u) 
res19: String = "1770b34b-4624-405f-85c1-04198b078586"
@  

```

Any ideas?

natechols:
For this to be most useful I think it needs to come with a built-in method - but we could name it something other than `toString` if there's a good reason not to override it.

mpkocher:
Would `toUnboxedString` work? It's a bit verbose, but I'm blanking on a name for an alternative toString.

natechols:
or `toIdString`, or `toSimpleString`, ... I don't have strong preferences.

mpkocher:
@natechols Does this work for you?

natechols:
LGTM

PR #284, opened 2016-08-10
More client testing and cleanup

natechols:


smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
```diff
@@ -69,15 +72,15 @@ object PbServiceParser {
   }
 
   // is there a cleaner way to do this?
-  private def entityIdOrUuid(entityId: String): Either[Int, UUID] = {
+  private def entityIdOrUuid(entityId: String): IdAble = {
```
mpkocher: (Not tested)
```
def entityIdOrUuid(sx: String): Try[IdAble] = {
    Try {sx.toInt}.map(IntIdAble(_)) match {
      case Success(x) => Success(x)
      case Failure(_) => Try {UUID.fromString(sx)}.map(UUIDIdAble(_))
    }
  }
```

PR #291, opened 2016-08-11
Tidy/cleanup suggestions to Pbservice

mpkocher:


natechols:
LGTM.  I guess CircleCI is having an off week...

mpkocher:
I'm closing this. IntelliJ is introducing a few too many diffs that aren't related. 

I'm going to take another pass at this to introduce a common pattern to remove some duplication.

The main pattern is Try {} then map/flatMap to compose one or more service operations. A for comprehension makes this is easy to understand the data flow. Then, emit to console, or write a file. Then map Try[T] to an exit code. 

```scala
def tryToInt[T](t: Try[T]): Int = {
  t match {
    case Success(_) => 0
    case Failure(ex) => println("Error ${ex.getMessage}")
  }
}
```

Or something along these lines. This would remove a bit of duplication.

PR #294, opened 2016-08-11
Fix config service error responses

smcclellan7:


mpkocher:
Please remove the *.db file.

smcclellan7:
Done.

PR #295, opened 2016-08-12
start on pipeline-related testing

natechols:


smrt-server-sim/src/main/scala/com/pacbio/simulator/scenarios/StressTestScenario.scala
```diff
@@ -0,0 +1,96 @@
...
+    fail("Can't get SMRT server status") IF jobStatus !=? EXIT_SUCCESS,
+    jobId := ImportDataSet(reference, Var(FileTypes.DS_REFERENCE.fileTypeId)),
+    jobStatus := WaitForJob(jobId),
+    fail("Import job failed") IF jobStatus !=? EXIT_SUCCESS
+  )
+  // submit multiple jobs in quick succession, and make sure they all finish.
+  val pbsmrtpipeJobTests = (for (i <- 0 to nJobs) yield {
+      Seq(jobIds(i) := RunAnalysisPipeline(pipelineOpts))
+    }).flatMap(s => s) ++ (for (i <- 0 to nJobs) yield {
+      Seq(jobStatus := WaitForJob(jobIds(i), Var(maxTime)),
```
natechols: @smcclellan7 I think I'm doing this wrong - it gets stuck here long after the first job has already finished.  Is it actually possible to collect job IDs the way I'm attempting to do here?

natechols: After further experimentation I think my code is correct, but the use of an unordered Map to store the runnable jobs means they're not run in the order they were submitted.  Will try to fix that problem.

natechols:
When the runnableJobs Map is changed to LinkedHashMap, this works exactly as intended.

https://github.com/PacificBiosciences/smrtflow/pull/296

PR #296, opened 2016-08-12
better handling of quick vs. regular jobs

natechols:
fixes bug where excess jobs got stuck in CREATED state

### DIFF
--- a/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
+++ b/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
@@ -203,7 +203,7 @@ trait JobDataStore extends JobEngineDaoComponent with LazyLogging {
   val resolver: JobResourceResolver
   // This is local queue of the Runnable Job instances. Once they're turned into an EngineJob, and submitted, it
   // should be deleted. This should probably just be stored as a json blob in the database.
-  var _runnableJobs: mutable.Map[UUID, RunnableJobWithId]
+  var _runnableJobs: mutable.LinkedHashMap[UUID, RunnableJobWithId]
 
   /**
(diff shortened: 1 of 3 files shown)


smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
```diff
@@ -264,7 +266,25 @@ trait JobDataStore extends JobEngineDaoComponent with LazyLogging {
 
   override def getNextRunnableJobWithId: Future[Either[NoAvailableWorkError, RunnableJobWithId]] = {
```
mpkocher: You might want to consider a model that uses a filter by job type id to remove some duplication.

```scala
private def getNextRunnableJobByType(jobTypeFilter: JobTypeId => Boolean): Future[Either[NoAvailableWorkError, RunnableJobWithId]] = {
    val noWork = NoAvailableWorkError("No Available work to run.")
    _runnableJobs.values.find(j => j.state == AnalysisJobStates.CREATED && jobTypeFilter(j.job.jobOptions.toJob.jobTypeId)) match {
      case Some(job) =>
        getJobById(job.id).map {
          case Some(j) =>
            _runnableJobs.remove(j.uuid)
            Right(RunnableJobWithId(job.id, job.job, j.state))
          case None => Left(noWork)
        }
      case None => Future(Left(noWork))
    }
  }

  def filterByQuickjobType(j: JobTypeId): Boolean = QUICK_TASK_IDS contains j 
  def getNextRunnableQuickJobWithId = getNextRunnableJobByType(filterByQuickjobType)
  override def getNextRunnableJobWithId = getNextRunnableJobByType((j: JobTypeId) => !filterByQuickjobType
```

Misc note: the mutation of the `_runnableJobs.remove` is occurring within a Future

PR #301, opened 2016-08-17
nest authenticate within http method directives to enable unauthentic…

skinner:
…ated OPTIONS handling for preflights


mpkocher:
@skinner Looks like there a test failing with `'OWNER' is not equal to 'Owner'`.

PR #303, opened 2016-08-18
match RSII movie name format when finding metadata files recursively

natechols:
@mpkocher this look right to you?

### DIFF
--- a/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
+++ b/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
@@ -17,7 +17,6 @@ import spray.httpx
 import spray.json._
 import spray.httpx.SprayJsonSupport
 
-
 import scala.collection.mutable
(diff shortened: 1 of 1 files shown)


smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
```diff
@@ -307,7 +307,14 @@ class PbService (val sal: AnalysisServiceAccessLayer,
   private lazy val defaultPresets = PipelineTemplatePreset("default", "any",
     Seq[PipelineBaseOption](),
     Seq[PipelineBaseOption]())
+  private lazy val rsMovieName = """m([0-9]{6})_([0-9a-z]{5,})_([0-9a-z]{5,})_c([0-9]{16,})_(\w\d)_(\w\d)""".r
 
+  private def matchRsMovieName(file: File): Boolean = {
+    rsMovieName.findPrefixMatchOf(file.getName) match {
```
mpkocher: I think `rsMovieName.findPrefixMatchOf(file.getName).isDefined` will do the same thing.

PR #309, opened 2016-08-23
start adding projects support to pbservice

natechols:
@mpkocher @skinner any thoughts?  it's messy but it works (minus authentication of course).  I thought about making it possible to set a project for datasets generated by other job types but this will increase the complexity considerably and I'm not sure how useful it is.

### DIFF
--- a/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
+++ b/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
@@ -330,9 +330,9 @@ class PbService (val sal: AnalysisServiceAccessLayer,
     rsMovieName.findPrefixMatchOf(file.getName).isDefined
 
   // FIXME this is crude
-  protected def errorExit(msg: String): Int = {
+  protected def errorExit(msg: String, exitCode: Int = -1): Int = {
     println(msg)
-    1
+    exitCode
   }
 
   protected def printMsg(msg: String): Int = {
@@ -551,8 +551,7 @@ class PbService (val sal: AnalysisServiceAccessLayer,
     }
   }
 
-  def runImportDataSetSafe(path: Path,
-                           projectName: Option[String] = None): Int = {
+  def runImportDataSetSafe(path: Path, projectId: Int = 0): Int = {
     val dsUuid = dsUuidFromPath(path)
     println(s"UUID: ${dsUuid.toString}")
     Try { Await.result(sal.getDataSet(dsUuid), TIMEOUT) } match {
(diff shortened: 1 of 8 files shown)


smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
```diff
@@ -659,6 +686,56 @@ class PbService (val sal: AnalysisServiceAccessLayer,
...
+              Await.result(sal.updateProject(projectId, request), TIMEOUT)
+            } match {
+              case Success(p) =>
+                if (verbose) printProjectInfo(p) else printMsg(s"Added dataset to project ${p.name}")
+              case Failure(err) => errorExit(s"Couldn't add dataset to project: ${err.getMessage}")
+            }
+          case Failure(err) => errorExit(s"Couldn't retrieve dataset $dsId")
+        }
+      case Failure(err) => errorExit(s"Couldn't retrieve project $projectId: ${err.getMessage}")
+    }
```
mpkocher: If possible, we should try to use map/flatMap/for comprehensions to compose the computation.

```scala
def addDataSetToProject(dsId: IdAble, projectId: Int, verbose: Boolean = false): Int = {
  val tx = for {
    project <- Try { Await.result(sal.getProject(projectId), TIMEOUT) }
    ds <- Try { Await.result(sal.getDataSet(dsId), TIMEOUT) }
    request <- Try { project.asRequest.appendDataSet(ds.id) }
    projectWithDataSets <- Try { Await.result(sal.updateProject(projectId, request), TIMEOUT) }
  } yield projectWithDataSets
  
  tx match {
    case Success(p) => if (verbose) printProjectInfo(p) else printMsg(s"Added dataset to project ${p.name}"); 0
    case Failure(ex) => errorExit(s"Failed to Process $projectId with $dsId")
  }
}
```

While this doesn't return specific messages, a util func could be added to add a context specific error message.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/models/Models.scala
```diff
@@ -547,7 +556,16 @@ case class ProjectRequest(
...
-    members: Option[Seq[ProjectRequestUser]])
+    members: Option[Seq[ProjectRequestUser]]) {
+
+  // this returns a copy!
+  def appendDataSet(dsId: Int): ProjectRequest = {
+    this.copy(datasets = datasets match {
+      case Some(ds) => Some(ds ++ Seq(RequestId(dsId)))
+      case None => Some(Seq(RequestId(dsId)))
+    })
+  }
```
mpkocher:
```scala
def appendDataSet(dsId: Int): ProjectRequest = {
  val datasets = datasets.map(ds => ds ++ Seq(RequestId(dsId)).getOrElse(Seq(RequestId(dsId))
  this.copy(datasets = datasets)
}
```

PR #312, opened 2016-08-24
Try refactoring to PbService

mpkocher:


natechols:
Very nice.  So in the context of a refactored PbService.scala, I could then do something like this in the various run methods instead of calling `waitForJob`, correct?

`for {
   job <- sal.pollForJob(jobId)
  ...
 } yield whatever
`

mpkocher:
That's the general idea. You can compose those units of computation via the Try monad. 

I didn't realize how much that initial method infected the entire model. I started making more changes, but this seemed to require changing the major of calls in client layer.

Some considerations:

Adding:

```scala
def TryToInt[T](msg: String, t: Try[T]): Int = {
    t match {
      case Success(_) => 0
      case Failure(ex) => errorExit("$msg ${ex.getMessage}")
    }
  }
```

Enables changing 

```scala
def runStatus(asJson: Boolean = false): Try[Status] =
    Try { Await.result(sal.getStatus, TIMEOUT) }.map(status => printStatus(status, asJson); status))
```

And in

```scala
case Modes.STATUS => TryToInt[Status](s"Failed to get status from $host;$port", case Modes.STATUS => ps.runStatus(c.asJson))
```

There's probably a cleaner way to do this, but this model might help remove some duplication.

PR #314, opened 2016-08-25
Dataset export (WIP)

natechols:
Starting with just the core functionality so I can get feedback, but I'm going to keep pushing additional layers as they get written.


smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/datasets/io/DataSetExport.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import com.typesafe.scalalogging.LazyLogging
+import collection.JavaConversions._
+import collection.JavaConverters._
+
+import com.pacbio.secondary.analysis.datasets._
+import com.pacificbiosciences.pacbiobasedatamodel.InputOutputDataType
+
+
+class ExportDataSets(datasets: Seq[Path], dsType: String, zipPath: Path)
```
mpkocher: Can you change this to `dsType: DataSetMetaTypes.DataSetType`?

smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/datasets/io/DataSetExport.scala
```diff
@@ -0,0 +1,90 @@
...
+            if (extRes2 == null) 0 else extRes2.getExternalResource.map { rr =>
+              writeResourceFile(dsId, rr, basePath) + {
+                val fi = rr.getFileIndices
+                if (fi == null) 0 else fi.getFileIndex.map {
+                  writeResourceFile(dsId, _, basePath)
+                }.sum
+              }
+            }.sum
+          } + {
+            val fi = er.getFileIndices
```
mpkocher: To handle null cases from the XSD based class, please use Option.

```scala
Option(er.getFileIndices).map(writeResourceFile(dsId, _, basePath))
```

PR #330, opened 2016-09-13
fix typo in create_roles.sh

skinner:


skinner:
This is trivial so I'm going to go ahead and merge

PR #331, opened 2016-09-13
Change PbBioinformatics role to PbBioinformatician, add run-qc scope to PbBioinformatician role

smcclellan7:
- Change PbBioinformatics role to PbBioinformatician
- Add run-qc scope to PbBioinformatician role
- Create api-manager.xml for WSO2 that enables passing JWTs to the backend.  (The bundler needs to replace the install default at {wso2-server-home}/repository/conf/api-manager.xml)
- Replace the "jwtAuth" spray auth magnet with "wso2Auth" to extract the username and roles claims from the WSO2 JWT without validating the signature.
- Eliminate the UserDao and UserService.


smcclellan7:
@skinner do you have thoughts on this PR?

skinner:
Thanks for doing this.  If we can just drop this in and use AM's JWTs, that's great.

I think api-manager.xml should go into the bundler; I added some wso2 files to the bundler in this change:
http://swarm/changes/185400

Hopefully that's an example of what you need to do to add a WSO2 config file to the bundler.

It looks like there are a bunch of relevant test failures.

I assume there's some way to configure the mapping between LDAP entries and the wso2 JWT claims?  That can come later, though.

mpkocher:
@skinner I believe this is what you're referring to:

https://docs.wso2.com/display/AM200/Passing+Enduser+Attributes+to+the+Backend+Using+JWT

It looks like the default claim implementation will have the "resource owner name" and will require us to extend that interface to the get the required fields that we need. 

Also, I'm a bit curious about the usage of Roles in the Claim.

smrt-server-base/src/main/scala/com/pacbio/common/auth/JwtUtils.scala
```diff
@@ -38,6 +45,32 @@ trait JwtUtilsProvider {
...
+      un <- Try(eu.asInstanceOf[String])
+    } yield ClaimsSet( // TODO(smcclellan): Use real values here, or make claims optional
+      id = un,
+      userName = un,
+      email = "",
+      firstName = "",
+      lastName = "",
+      iat = new JodaDateTime,
+      exp = new JodaDateTime,
+      roles = Set.empty,
```
mpkocher: Can you comment on the use of roles here?  I.e., should this have scopes and roles? Or just scopes?

mpkocher:
Can we split this into two PRs? 

The first one upto https://github.com/PacificBiosciences/smrtflow/pull/331/commits/b0a3a7b44e9404fe3a78f3ef9f78649726594990 and a second PR to have claim and JWT parsing updates from the services.

smrt-server-base/src/main/scala/com/pacbio/common/auth/JwtUtils.scala
smcclellan7: The scopes are not part of the standard WSO2 claims set, nor, I would argue, should they be.  The point of scopes, as I understand them, is to speak to the UI about permissions using UI-comprehensible language.  That's why the scopes are named after the the UI panels.  They're merely a prism, by whose refracted colors the ignorant UI may view the light of truth, the light we call "roles."  To the extent that SMRTLink cares about permissions at all, and that extent should be as small as possible, I think it should consider the roles, rather than trying to parse a translation of the roles that's meant for the UI.

PR #338, opened 2016-09-21
Rename analysis scope

skinner:


skinner:
not sure about the admin scope; it didn't seem like it was working today

PR #340, opened 2016-09-22
Api manager fixes

skinner:
Some fixes to API manager issues from yesterday and today.  The analysis scope name changes fixes the job creation failures from yesterday, and the x-auth-type and content-disposition change fix the download failures.

### DIFF
--- a/extras/wso2/smrtlink_swagger.json
+++ b/extras/wso2/smrtlink_swagger.json
@@ -1014,7 +1014,7 @@
     },
     "/secondary-analysis/datastore-files/{uuid}/download": {
       "get": {
-        "x-auth-type": "Application%20%26%20Application%20User",
+        "x-auth-type": "None",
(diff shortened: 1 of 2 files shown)


extras/wso2/smrtlink_swagger.json
```diff
@@ -31,8 +31,8 @@
           "description": "Data%20Management%20Authorization"
         },
         {
-          "name": "smrt-analysis",
-          "key": "smrt-analysis",
+          "name": "analysis",
+          "key": "analysis",
```
mpkocher: I believe this would probably be better as `smrt-analysis` to be consistent with the other scopes  and the companion panel names (e.g., run-design -> Run Design). But we can deal with this in the next iteration.

skinner: It originally came from the silo names in the code on the client.

PR #344, opened 2016-09-27
Add iui-basic scope

smcclellan7:


mpkocher:
Not really sure I understand the motivation for adding this.

PR #345, opened 2016-09-27
Add info to root level and add description to all response 

mpkocher:
These changes make the swagger file valid.

Using node_modules/swagger-cli/bin/swagger.js validate /path/to/swagger.json

Can be installed npm install swagger-cli


mpkocher:
@smcclellan7 does this work for you?  Programmatically adding those keys created a pretty noisy diff.

I think we should take another pass and convert that useage of urlencoding values to non-urlencoded values. I'm a bit confused how that even works within wso2.

PR #347, opened 2016-09-28
Add ScopeService

smcclellan7:
It seems that WSO2 ignores scope definitions that are not associated with any endpoint.  This is a problem for the 'iui-basic' scope that is not supposed to be associated with any SMRT Link endpoint, but simply used as an indicator of whether the user should be allowed to log into the Insrument UI.

By adding this new service, we can ensure that every scope is associated with an endpoint, and it gives a simple way to check whether a given access token allows access to a given scope.  (As a bonus, I threw in similar endpoints to check whether an access token allows access to a given role.)

I don't know if this is necessarily how we want to solve this problem, but it's an idea.  I'd love to hear thoughts?

### DIFF
--- a/extras/wso2/smrtlink_swagger.json
+++ b/extras/wso2/smrtlink_swagger.json
@@ -1231,6 +1231,138 @@
         "x-auth-type": "Application%20%26%20Application%20User"
       }
     },
+    "/scope/run-design": {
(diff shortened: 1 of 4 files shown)


mpkocher:
We decided to approach the problem differently.

PR #348, opened 2016-09-28
Fix urlencoded values in swagger.json

mpkocher:


smcclellan7:
Does this work?  When I tried using a file like this, without the urlencoding, on a local WSO2 setup on my computer it failed.

But I'm using the extras/wso2/api_create.sh script, is the bundler using something else to create the API?

PR #351, opened 2016-10-05
Add route for pbsmrtpipe job termination to swagger file

mpkocher:


mpkocher:
@skinner Not a whole lot going on here. I'm going to merge.

PR #352, opened 2016-10-07
allow setting pbservice host/port defaults with env vars

natechols:


smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/PbService.scala
```diff
@@ -138,11 +140,11 @@ object PbServiceParser {
 
     opt[String]("host") action { (x, c) =>
       c.copy(host = x)
-    } text "Hostname of smrtlink server"
+    } text "Hostname of smrtlink server.  Override the default with env PB_SERVICE_HOST (default: localhost)"
```
mpkocher: Perhaps having the default value would be useful `s"(default: $defaultHost)"` ?

natechols: But what happens when the environment variables are already set and the user runs `pbservice --help`?

mpkocher: I view the ENV vars or if pbservice could load an RC config file from ~/.smrtlink.json as overriding default values. This seems like the expected behavior to display the actual default values that should be used. Do you have a different suggested approach? 

Here's the python version

```
$> export PB_SERVICE_HOST="smrtlink-test"
$> pbservice status --help
usage: pbservice status [-h] [--host HOST] [--port PORT] [--log-file LOG_FILE]
                        [--log-level {DEBUG,INFO,WARNING,ERROR,CRITICAL}]
                        [--debug] [--quiet]

optional arguments:
  -h, --help            show this help message and exit
  --host HOST           Server host. Override the default with env
                        PB_SERVICE_HOST (default: smrtlink-test)
  --port PORT           Server Port. Override default with env PB_SERVICE_PORT
                        (default: 8070)
  --log-file LOG_FILE   Write the log to file. Default(None) will write to
                        stdout. (default: None)
  --log-level {DEBUG,INFO,WARNING,ERROR,CRITICAL}
                        Set log level (default: INFO)
  --debug               Alias for setting log level to DEBUG (default: False)
  --quiet               Alias for setting log level to CRITICAL to suppress
                        output. (default: False)
```

PR #356, opened 2016-10-15
Add version to status. Add Manifest to Client

mpkocher:
Add Service to get Manifest by Id. Update Client and get-manifest and get-manifests to pbservice.

- get-manifest will return "smrtlink" if a manifest id is not provided and has some level of backward compatibility (i.e, doesn't use the new service that enables getting a component by id)


mpkocher:
@natechols Can you review this?

PR #358, opened 2016-10-18
Create UserService

smcclellan7:


smrt-server-base/src/main/scala/com/pacbio/common/models/Models.scala
```diff
@@ -136,15 +136,57 @@ object Roles {
   object PbLabTech extends Role { override val name = "Internal/PbLabTech" }
   object PbBioinformatician extends Role { override val name = "Internal/PbBioinformatician" }
 
-  def fromString(name: String): Option[Role] =
-    Seq(PbAdmin, PbLabTech, PbBioinformatician).find(_.name == name)
-}
+  val ALL_ROLES = Set(PbAdmin, PbLabTech, PbBioinformatician)
```
mpkocher: Can you clarify what these are required?

smrt-server-base/src/main/scala/com/pacbio/common/services/HealthService.scala
```diff
@@ -51,7 +51,7 @@ class HealthService(healthDao: HealthDao, authenticator: Authenticator)
             } ~
             post {
               entity(as[HealthMetricCreateMessage]) { m =>
-                authorize(authInfo.hasPermission(PbAdmin)) {
+                authorize(user.hasPermission(PbAdmin)) {
```
mpkocher: Can this be removed since wso2 is handling authorization?

smrt-server-base/src/main/scala/com/pacbio/common/models/Models.scala
smcclellan7: They're not.  Removed.

smrt-server-base/src/main/scala/com/pacbio/common/services/HealthService.scala
smcclellan7: Yep.

smrt-server-base/src/main/scala/com/pacbio/common/models/Models.scala
```diff
@@ -136,15 +136,36 @@ object Roles {
...
 
-object UserRecord {
-  def apply(userName: String, roles: Roles.Role*): UserRecord = UserRecord(userName, Set(roles:_*))
+  def fromString(name: String): Option[Role] = ALL_ROLES.find(_.name == name)
 }
 
-case class UserRecord(userName: String, roles: Set[Roles.Role] = Set.empty)
+case class UserRecord(userName: String,
```
mpkocher: Minor quibble. I think you're using "username" in the wso2 context, which means user id? Having userName, firstName and lastName is a bit unclear. I would have guessed that userName would be first+last name, but this is getDisplayName. Is it worth it to switch to `userId` or does this create an unnecessary mapping layer to map to wso2 naming model.

smcclellan7: Done:
userName -> userId
Role type -> String

PR #359, opened 2016-10-19
Add some basic docs on CLI tools

mpkocher:


mpkocher:
@natechols This is still a bit rough, but it's a starting point. 

Not sure what to do with those out of data Endpoint docs. 

I believe we should use https://github.com/Arello-Mobile/swagger2rst within readthedocs to generate the rst from the swagger.json, then add those RST to the current sphinx project in smrtflow.

PR #362, opened 2016-10-21
Initial Checking of pbbundler

mpkocher:
I tried to only import the minimal code from p4


mpkocher:
To make this work a few changes would have be made

- Explicitly pass the path to the tomcat zip to the fabric tasks
- update the jenkins*.sh to use the code in smrtflow
- (optional) remove the LDAP_* from config.json

mpkocher:
@natechols Does this look like a reasonable starting point?

mpkocher:
I want to get this working in the mainline build. I will delete all the LDAP related options in a different PR.

PR #364, opened 2016-10-24
make pbbundler easier to run outside of jenkins job

natechols:
@mpkocher I can now build this locally, I think it'll require a few minor tweaks to Jenkins to make it work there.


mpkocher:
Look good. Thanks.

PR #371, opened 2016-11-01
Bugfix/itg 220 internal tools website does not recognize

evolvedmicrobe:


evolvedmicrobe:
From ITG-220, moving from bitbucket to github.  @mpkocher could I get this reviewed?

mpkocher:
@evolvedmicrobe I don't really think this makes a lot of sense to start hardcoding these special values. For ports other than the default, 8081, the port needs to be explicitly provided as "<host>:<port> (e.g, smrtlink-my-host:9321") in the Condition CSV. This should probably be documented somewhere, perhaps in the UI?

evolvedmicrobe:
What's the disadvantage of this? The problem with documentation is someone has to write it, someone else has to read it, and then somebody has to keep it up to date.  We haven't had a great track record of that, and we're already hardcoding the 8081 port.  This solution (or moving SMRTLink-Release back to the default) seems easiest to me.

evolvedmicrobe:
So, okay to merge?

mpkocher:
@evolvedmicrobe I would discourage adding hardcoded host names in compiled code if possible.

The current Reseq Condition CSV API supports "<host>:<port>" or a "<host>" where the default port will be set to 8081. Unfortunately, this doesn't appear to be documented. 

https://github.com/PacificBiosciences/smrtflow/tree/master/smrt-server-analysis-internal#csv-example-file-and-services

Adding a hardcoded lookup table of SL hosts to resolve default ports seems like not the greatest idea. It also makes documenting the interface a bit ungraceful.

> The SL host can be provided as "<host"> or "<host>:<port">. If the port isn't explicitly provided, then the 8081, however if the host is "smrtlink-release", then the port will default to 9091. 

This suggest that the handling of the port should be removed and the hardcoded SL host->port registry should complete handle this, or the internal resolving or the port should be removed. It's also unclear where future changes should be made. If a user adds a new SL server on 7777, should they update the scala code or be required to use the CSV input of smrtlink-my-server:7777 ? 

Note that the internal default port is 8080/8081, however, the customer systems have a default port of 9090/9091 from the SL installer. Hence, other internal systems might have this "non-standard" port and require updating the scala code. 

If you wanted some sugar to avoid explicitly specifying the port, I would have preferred for resolving the SL port layer to be sorted out at the client side (the UI code). Perhaps from an explicit registry/map of SL hosts in a JSON file along side the UI code to avoid having to recompile the JS react code. 

Nevertheless, I don't want to spend any more time on this. Please update the docs to communicate the interface changes of the Reseq Condition CSV file and merge. The docs are here:

https://github.com/PacificBiosciences/smrtflow/tree/master/smrt-server-analysis-internal#csv-example-file-and-services

Please create a new issue in the SE or ITG backlog to remove the `smrt-server-analysis-internal` subproject from the smrtflow repo and into an ITG owned space within BB.  You'll probably need to update the necessary build scripts as well. This will enable you to make whatever changes you think believe are acceptable in the future.

evolvedmicrobe:
@mpkocher Just so I am 100% crystal clear on this before closing this pull request - An actual user has complained about a problem that this would resolve, but you do not want to make this change because although you can't identify any user today who would be negatively affected by it, you feel that the change is unhygienic to the code base, and would possibly create negative user experiences in a hypothetical future?

PR #372, opened 2016-11-02
started simulator testing of delete jobs, plus pbservice compatibility

natechols:
@mpkocher had to change `isActive` to `Optional[Boolean]` to keep pbservice from choking on older smrtlink instances

### DIFF
--- a/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobs/JobModels.scala
+++ b/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobs/JobModels.scala
@@ -128,7 +128,7 @@ object JobModels {
       createdBy: Option[String],
       smrtlinkVersion: Option[String],
       smrtlinkToolsVersion: Option[String],
-      isActive: Boolean = true) {
+      isActive: Option[Boolean] = Some(true)) {
 
       def isComplete: Boolean = AnalysisJobStates.isCompleted(this.state)
       def isSuccessful: Boolean = this.state == AnalysisJobStates.SUCCESSFUL
@@ -148,7 +148,7 @@ object JobModels {
           createdBy: Option[String],
           smrtlinkVersion: Option[String],
           smrtlinkToolsVersion: Option[String],
-          isActive: Boolean = true) = {
+          isActive: Option[Boolean] = Some(true)) = {
 
           // This might not be the best idea.
           val state = AnalysisJobStates.intToState(stateId) getOrElse AnalysisJobStates.UNKNOWN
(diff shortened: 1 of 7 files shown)


mpkocher:
Is it possible to sort this out at the IO layer in the JSON protocol layer and default to true if isActive is not there? Currently, this backward compatibility motivation is resulting in database changes to support this, which doesn't sound right. 

The migration doesn't look consistent with the new definitions?

https://github.com/natechols/smrtflow/blob/ad3c4f0a8cfd9aaacd82deaae60ea100b2aba073/smrt-server-link/src/main/scala/db/migration/V19__JobDelete.scala#L15

https://github.com/natechols/smrtflow/blob/ad3c4f0a8cfd9aaacd82deaae60ea100b2aba073/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/database/TableModels.scala#L101

Is sqlite's type looseness the reason why there aren't errors?

It's also not clear what `isActive=None` is intended to mean.

mpkocher:
I'm not sure what the default behavior of sqlite is, but I think one of these needs to have an explicit not null to be consistent with the slick table definition?

https://github.com/natechols/smrtflow/blob/ad3c4f0a8cfd9aaacd82deaae60ea100b2aba073/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/database/TableModels.scala#L357

https://github.com/natechols/smrtflow/blob/ad3c4f0a8cfd9aaacd82deaae60ea100b2aba073/smrt-server-link/src/main/scala/db/migration/V19__JobDelete.scala#L16

natechols:
Yes, we can sort this out at the IO layer, but that means writing a custom JSON protocol for the affected models.  I wasn't sure which model we preferred - the custom protocol will be gross but it would at least keep the rest of the code simpler, and we could potentially remove it in the future.  Should I take that approach?

My naive assumption was that using `Option[Boolean]` simply means that NULL values will be allowed.  This does seem like black magic to me.

mpkocher:
I agree that this is not awesome to have to write custom spray JSON protocols. However, this might isolate the required changes. 

You bring up a good point. The client really won't strictly work with older versions (i.e., pbservice delete-job will fail on smrtlink-beta). Not sure how to handle this. When adding a new endpoint or changes to the public API (e.g., data models), it might be good to bump the smrtflow version. Perhaps in future versions we could require the client to check /status to get the version and fail early. if it's not compatible.

natechols:
In principle I agree that we should enforce version compatibility in pbservice.  In practice, our internal users would freak out if we apply this universally.  Maybe just tag specific functionality like this as requiring a minimum version?

natechols:
I will start over with a new branch.

PR #376, opened 2016-11-03
Create SHOULD_RAISE operator for Sim

smcclellan7:


smrt-server-sim/src/main/scala/com/pacbio/simulator/steps/ConditionalSteps.scala
```diff
@@ -26,10 +28,27 @@ trait ConditionalSteps {
     }
   }
 
+  case class ExpectErrorStep(step: Step) extends Step {
+    override val name = s"ExpectError-${step.name}"
+    override def run: Future[Result] = step.run.recoverWith {
+      case NonFatal(ex) => Future { EXCEPTION(ex) }
```
mpkocher: Are there any concerns here with false positives?

smcclellan7: Like, do you want to filter on which exceptions are raised?  E.g., SHOULD_RAISE(IllegalArgumentException)

smcclellan7: PTAL

mpkocher:
Looks reasonable. It might be better to do this with partial functions versus reflection, but this can be addressed in a future iteration. I want to get this in and unblock testing efforts for the delete job for @natechols.

PR #377, opened 2016-11-03
more delete-job tests, including pbsmrtpipe job and failure mode

natechols:
@mpkocher is there anything else I need to include in these tests before I declare the issue closed?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,7 +12,7 @@
 
 name := "smrtflow"
 
-version in ThisBuild := "0.1.11-SNAPSHOT"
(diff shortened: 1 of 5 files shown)


mpkocher:
A few comments:

- update the schema migration to include not null (if the default is null)
- update the swagger file
- bump the smrtflow version to 0.2.0
- Add some [terse docs](https://github.com/PacificBiosciences/smrtflow/blob/master/smrt-server-analysis/README.md) until we get more moment behind the swagger file.

natechols:
Not sure what you're asking for in the docs.

PR #378, opened 2016-11-04
backwards compatibility for EngineJob unmarshalling

natechols:
@mpkocher @smcclellan7 please review; in the process I did a bunch of refactoring that I had thought was necessary at the time - this eliminates the conflict between our custom JSON protocols and allows us to use simple inheritance.  On the other hand, I don't think this refactoring is actually necessary to get my new protocol to work (changing `JsonFormat` to `RootJsonFormat` in `EngineJobProtocol` was what I really needed).  Your call whether we go with these changes or stick to the more limited solution.


natechols:
Screw it, the tests are broken anyway.  I'll go with the simpler fix.

PR #383, opened 2016-11-08
Validate project user role

smcclellan7:


mpkocher:
@skinner Can you review this?

skinner:
The role stuff looks good to me; I don't know the reason for the build.sbt changes, but if they make the tests work, I'm on board.

I've gone back and forth over whether I think it's important to have things like these roles as a sealed trait (checkable by the compiler) or as strings.  But this diff actually shows one place where I changed the strings and forgot to update the test, so I think that's an argument for the sealed trait.
👍

smcclellan7:
Personally, I'm a big fan.  I'm constantly misspelling things, or changing something in one place and forgetting to change it in another.  I like to be able to do the ctrl-b in my intellij and see all the uses of the case object.  But as in all things, I submit to the will of the majority.  #democracy

PR #384, opened 2016-11-08
add project deletion endpoint

skinner:


smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
```diff
@@ -151,6 +151,24 @@ trait ProjectDataStore extends LazyLogging {
     db.run(fullAction.transactionally)
   }
 
+  def deleteProjectById(projId: Int): Future[Option[Project]] = {
```
mpkocher: What's the motivation for having `Future[Option[Project]]`?

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/models/Models.scala
```diff
@@ -514,7 +514,8 @@ case class Project(
     description: String,
     state: String,
     createdAt: JodaDateTime,
-    updatedAt: JodaDateTime) {
+    updatedAt: JodaDateTime,
+    isActive: Boolean) {
```
mpkocher: Would you mind adding a terse comment on the meaning of `isActive` and how it should be interpreted?

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
skinner: I was just copying what Nat did for job deletion; I figured I might as well try to be consistent between job/project and it seemed like a potentially useful response.  I'm not attached to it, though, and there's no product requirement for it so far, so we could take it out.

natechols: To elaborate, I was mimicking `getJobBy*`, which returns `Future[Option[EngineJob]]`.

smrtlink_swagger.json
```diff
@@ -816,6 +816,30 @@
         "x-scope": "data-management",
         "x-auth-type": "Application & Application User",
         "x-throttling-tier": "Unlimited"
+      },
+      "delete": {
```
skinner: I added some basic description fields; I'm not sure if we want to get into the structure of the requests and responses in this swagger or have those externally

PR #387, opened 2016-11-10
backend support for EULAs

natechols:
@mpkocher please review, this is just the database layer but I would like to confirm that this is an appropriate approach before I pile too much on top


mpkocher:
Looks good. Can you update the swagger file and add a terse description of the new endpoint(s)?

smrt-server-link/src/main/scala/db/migration/V21__EulaRecord.scala
```diff
@@ -0,0 +1,34 @@
...
+import com.pacbio.common.time.PacBioDateTimeDatabaseFormat
+
+class V21__AddEula extends JdbcMigration with SlickMigration with LazyLogging {
+
+  override def slickMigrate(db: DatabaseDef): Future[Any] = db.run {
+    V21Schema.eulas.schema.create
+  }
+}
+
+object V21Schema extends PacBioDateTimeDatabaseFormat {
```
mpkocher: This looks good. It probably better if we try to consistently use one approach to these migrations. 

Consider adding a primary key on `smrtlinkVersion`?

```
def smrtlinkVersion: Rep[String] = column[String]("smrtlink_version", O.PrimaryKey)
```

PR #390, opened 2016-11-14
Configure circle.yml to test sbt projects separately.

smcclellan7:


natechols:
Minor nitpick: instead of going through the submodules alphabetically, it should do them in order of API hierarchy, i.e. smrt-server-base before smrt-server-link before smrt-server-analysis, etc.  Otherwise LGTM.

smcclellan7:
@natechols done.

smcclellan7:
Done. @mpkocher, PTAL

PR #392, opened 2016-11-22
add DELETE for EULA endpoint

natechols:


smrt-server-link/src/main/scala/db/migration/V22__AddOsVersionToEulaRecord.scala
```diff
@@ -0,0 +1,15 @@
...
+import org.flywaydb.core.api.migration.jdbc.JdbcMigration
+import slick.jdbc.JdbcBackend.DatabaseDef
+
+import slick.driver.SQLiteDriver.api._
+import slick.jdbc.JdbcBackend.DatabaseDef
+
+import scala.concurrent.Future
+
+class V22__AddOsVersionToEulaRecord extends JdbcMigration with SlickMigration {
+  override def slickMigrate(db: DatabaseDef): Future[Any] = db.run(DBIO.seq(
```
skinner: DBIO.seq should only be necessary when there's more than one DBIO.  IIRC the sqlu creates a DBIO object; I think you can pass that straight to db.run.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
```diff
@@ -1189,14 +1192,24 @@ trait DataSetStore extends DataStoreComponent with LazyLogging {
                         enableInstallMetrics: Boolean,
                         enableJobMetrics: Boolean): Future[EulaRecord] = {
     val now = JodaDateTime.now()
-    val rec = EulaRecord(user, now, smrtlinkVersion, enableInstallMetrics, enableJobMetrics)
+    // FIXME this should probably move somewhere central
+    val osVersion = if (Files.exists(Paths.get("/proc/version"))) {
```
mpkocher: Is there not a java API todo this? Similar to `platform` in python?

extras/pbbundler/smrtlink_services_ui/bin/apply-config
```diff
@@ -529,7 +537,8 @@ def to_smrt_link_ui_config_d(host, services_port, smrtview_port, service_event_u
     d = {"smrt-view": {"default-server": smrtview_uri},
          "smrt-link": {"default-server": services_uri},
          "auth": {"default-server": auth_uri},
-         "events-url": {"default-server": service_event_url}
+         "events-url": {"default-server": service_event_url},
```
mpkocher: Instead of adding URLs for specific cases, did you consider a single interface that supports a list of URLs as `events-url` endpoints (e.g., `events-urls:{default-server:["http://a.com/events", "https://b.com/my-events"]}`

Is this going to require a adding an Auth token that must be sent for the Tech Support URL?

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/actors/JobsDao.scala
natechols: To the best of my knowledge, the fallback approach here *is* the usual java API (or the apache-commons wrapper thereof), but based on my experiments it's clearly an inferior solution.

extras/pbbundler/smrtlink_services_ui/bin/apply-config
natechols: I don't know how this is supposed to work in production - I'm pretty sure it's TBD, so there is nothing stopping us from changing this around, but we needed a proof-of-concept working in a hurry.

PR #393, opened 2016-11-28
handle spaces in path correctly in files service

natechols:
@mpkocher pretty sure i did this right but someone should double-check that there won't be side effects

### DIFF
--- a/smrt-server-base/src/main/scala/com/pacbio/common/services/AbstractFilesService.scala
+++ b/smrt-server-base/src/main/scala/com/pacbio/common/services/AbstractFilesService.scala
@@ -1,6 +1,7 @@
 package com.pacbio.common.services
 
 import java.io.File
+import java.net.URLDecoder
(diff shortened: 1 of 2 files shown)


skinner:
LGTM.  The test is nice and clean, thanks.

mpkocher:
@natechols Looks good. However, there might be other places where this breaks.

https://github.com/PacificBiosciences/smrtflow/blob/master/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/pbsmrtpipe/IOUtils.scala#L142

I think these calls with DataSets as entry points to pbsmrtpipe might have to be quoted correctly to support spaces? On the pbsmrtpipe side, I think it might handle this correctly in downstream tasks if the entry points are correctly quoted.

natechols:
Good point, the pbsmrtpipe call will probably break.  (pbsmrtpipe itself does appear to work as long as the input args are quoted, however.)

PR #394, opened 2016-11-29
another spaces fix

natechols:


smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/pbsmrtpipe/IOUtils.scala
```diff
@@ -139,7 +139,7 @@ object IOUtils extends LazyLogging{
       workflowOptions: Seq[PipelineBaseOption],
       serviceUri: Option[URI]) = {
 
-    val e = entryPoints.map(x => s"-e ${x.entryId}:${x.path.toString}").fold("")((a, b) => s"$a $b")
+    val e = entryPoints.map(x => s"""-e ${x.entryId}:"${x.path.toString}"""").fold("")((a, b) => s"$a $b")
```
mpkocher: Might want to consider using single quotes.

```
-e 'e_01:/path/to/file 01.xml'
```

PR #398, opened 2016-12-01
[SL-431] set backend port

skinner:
This sets the backend port and the swagger at services start time, fixing SL-431 and relieving us from having to update the wso2 zip when we add new endpoints to the swagger.

### DIFF
--- a/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/AmClient.scala
+++ b/smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/AmClient.scala
@@ -8,11 +8,15 @@ import scala.concurrent.{Await, Future}
 import scala.concurrent.duration._
 
 import akka.actor.ActorSystem
+import akka.util.Timeout
 
 import spray.json._
 
+import com.typesafe.config.ConfigFactory
+
 import scopt.OptionParser
 
+import com.pacbio.common.dependency.TypesafeSingletonReader
 import com.pacbio.logging.{LoggerConfig, LoggerOptions}
 
 import org.wso2.carbon.apimgt.rest.api.publisher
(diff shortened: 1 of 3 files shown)


smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/AmClient.scala
```diff
@@ -55,14 +59,23 @@ object AmClientParser {
   // Examples:
   // smrt-server-analysis/target/pack/bin/amclient set-api --target http://localhost:8090/ --swagger-resource /smrtlink_swagger.json --app-config ~/p4/ui/main/apps/smrt-link/src/app-config.json --user admin --pass admin --host login14-biofx02 --port-offset 10
 
+  val conf = ConfigFactory.load()
+  val target = if (conf.hasPath("pb-services.host") && conf.hasPath("pb-services.port")) {
```
mpkocher: This might be cleaner as a Try?

```scala
val targetx = for {
  host <- Try { conf.getString("pb-services.host") }
  port <- Try { config.getInt("pb-services.port")}
 } yield new URL(s"http://$host}:$port")
val target = targetx.toOption
```

skinner: Cool, done.

smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/AmClient.scala
```diff
@@ -55,14 +60,22 @@ object AmClientParser {
   // Examples:
   // smrt-server-analysis/target/pack/bin/amclient set-api --target http://localhost:8090/ --swagger-resource /smrtlink_swagger.json --app-config ~/p4/ui/main/apps/smrt-link/src/app-config.json --user admin --pass admin --host login14-biofx02 --port-offset 10
```
mpkocher: I would encourage trying to use specific types over generic strings if possible. For example, roles (Set[String]), target: URL. This would make the code a bit easier to digest.

skinner: I agree in general; in this case it's starting as a string (from the command line) and ending up as a string (in the endpoint config) without much processing in between, so I didn't think it was a big deal.  But I buy the readability argument for making it typed.  Done.

PR #402, opened 2016-12-05
Fix job datastore stderr

mpkocher:


mpkocher:
@skinner Can you review these changes? I've update the model to write errors to stdout (i.e., log) and now have the source id point to job stdout. The pbsmrtpipe job type is still not correct because the pbsmrtpipe log is considered the "master log". Nevertheless, this is a reasonable improvement.

skinner:
LGTM; having the master log be more of a log makes sense for those job types.  And I think leaving the pbsmrtpipe log as the master log for pbsmrtpipe jobs makes sense for now.

mpkocher:
@skinner it's worth noting that the pbsmrtpipe log won't have the error message from timeouts and that job type does NOT have parity with the other job types.

PR #404, opened 2016-12-07
[SL-619] Report migration errors and recreate job_events index

skinner:
We used to swallow migration exceptions in an Await.ready.  This changes the `ready` to a `result` so that exceptions propagate and get reported.  But that change means that the V17 migration that was generating an exception would stop the migration process.

So this also changes the V17 migration to only include the changes that were successfully made, before it hit the exception.  Changing a past migration that has been applied on various instances is something I would ordinarily say was a no-no, but I think this is a special case.  The effect of the new, shorter V17 is the same as the effect of the old V17; only the statement that resulted in an error (and the subsequent statements, which weren't getting run) have been removed.

It also re-adds the index on job_events.

### DIFF
--- a/smrt-server-link/src/main/scala/db/migration/SlickMigration.scala
+++ b/smrt-server-link/src/main/scala/db/migration/SlickMigration.scala
@@ -46,7 +46,7 @@ trait SlickMigration { self: JdbcMigration =>
   override final def migrate(conn: Connection): Unit = {
     val db = new UnmanagedDatabase(conn)
     try {
-      Await.ready(slickMigrate(db), Duration.Inf)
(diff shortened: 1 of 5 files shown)


skinner:
It turns out that the V8 migration was also broken, for I don't know how long.  Now I wonder whether deployed instances got the working V8 migration (successfully applied) or the broken V8 migration (which failed silently because the Await.ready was swallowing the exception).  But I don't think it matters, because a subsequent migration (V11_1) dropped and recreated the table that V8 modified.

skinner:
rebased to smrtanalysis-4.0.0 branch, please review and merge if approved

PR #407, opened 2016-12-08
WIP Add postgres to deps

mpkocher:


mpkocher:
The database subproject is no longer used and should be completely deleted, right?

smcclellan7:
@mpkocher - are there other changes you want to see, or testing you want to do, before we merge this?

mpkocher:
There are still some interfaces that need to be sorted out, otherwise mainline build will be broken.

- conversion tool from config.json 1.0 to config.json 2.0
- conversion tool from sqlite to postgresql 
- get run_sim_local.py to successfully run
- document config.json 2 schema

Both the stress.py and run_sim_local.py need to be migrated back to scala and run within sbt as part of the CI tests. There are too many configuration layers by calling these external integration tests. I  fixed some bugs with `run_sim_local.py` config setup, but it still doesn't pass.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/database/legacy/SqliteToPostgresConverter.scala
```diff
@@ -0,0 +1,643 @@
...
+import java.nio.file.{Paths, Path}
+import java.util.UUID
+
+import com.pacbio.common.time.PacBioDateTimeDatabaseFormat
+import com.pacbio.logging.{LoggerOptions, LoggerConfig}
+import com.pacbio.secondary.analysis.jobs.AnalysisJobStates
+import com.pacbio.secondary.analysis.jobs.JobModels.{EngineJob, JobEvent}
+import com.pacbio.secondary.analysis.tools.{ToolSuccess, ToolFailure, CommandLineToolRunner}
+import com.pacbio.secondary.smrtlink.database.{DatabaseConfig, DatabaseUtils, TableModels}
+import com.pacbio.secondary.smrtlink.models._
```
skinner: The dependency on the models (which can change) is unfortunate here.  What happens if they change during the rest of the iguana cycle?

We shouldn't have to convert to models during the data copying step.  If a row in the source database has the value 5 for an analysis job state, we can just copy the 5 over to the destination database without having to convert it into an analysis job state in between.

Ideally the data migration wouldn't depend on the models at all.  I've done it before with JDBC.  It's not immediately clear to me whether you can do it with slick (or how).  Worst case we could use something like https://github.com/swissmanu/JDBCCopier

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/database/legacy/SqliteToPostgresConverter.scala
```diff
@@ -0,0 +1,643 @@
...
+    e.printStackTrace(new PrintWriter(sw))
+    sw.toString
+  }
+}
+
+class PostgresWriter(db: slick.driver.PostgresDriver.api.Database, pgUsername: String) {
+  import slick.driver.PostgresDriver.api._
+  import TableModels._
+
+  case class FlywayRow(installed_rank: Int,
```
mpkocher: https://github.com/PacificBiosciences/smrtflow/blob/sqlite-to-postgres/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/database/DatabaseUtils.scala#L50

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/database/legacy/SqliteToPostgresConverter.scala
mpkocher: https://github.com/nafg/slick-migration-api

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/database/legacy/SqliteToPostgresConverter.scala
smcclellan7: Seems like a good idea.

smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/database/legacy/SqliteToPostgresConverter.scala
smcclellan7: I can take a gander at those libraries.

smrt-server-link/src/main/scala/db/migration/V1__InitialSchema.scala
```diff
@@ -1,490 +1,69 @@
...
 import slick.jdbc.JdbcBackend.DatabaseDef
-import slick.lifted.ProvenShape
 
 import scala.concurrent.Future
 
 // scalastyle:off
+/**
+  * First Postgresql Migration
+  *
+  * FIXME(mpkocher)(2016-12-9) copy Table models into here
```
mpkocher: Yes. Until we have to upgrade the first system that is not blown away be testing and we have to have db schema changes, this will require ripping the models dep out of there and creating new migrations. 

Similarly, any changes to models require updating the sqlite-to-postgres migration tool.

PR #409, opened 2016-12-12
Delete dataset service

natechols:
Including deleting entry points for an auto-merged dataset - I'm not sure if we really need to handle the recursive case.

### DIFF
--- a/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
+++ b/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
@@ -66,7 +66,7 @@ abstract class DeleteResourcesBase(opts: DeleteResourcesOptionsBase)
   }
 
   protected def toReport(targetPaths: Seq[Path], deletedFiles: Seq[DeletedFile]): Report = {
-    val nErrors = deletedFiles.count(_.wasDeleted == false).toLong
+    val nErrors = deletedFiles.count(_.wasDeleted == false)
     val nBytesTotal = deletedFiles.map(_.nBytes).sum
     val tables = if (deletedFiles.nonEmpty) {
       val paths = ReportTableColumn("path", Some("Path"), deletedFiles.map(_.path))
@@ -81,7 +81,7 @@ abstract class DeleteResourcesBase(opts: DeleteResourcesOptionsBase)
         ReportStrAttribute("job_dir", "Directory", targetPaths.mkString(", ")),
         ReportLongAttribute("n_errors", "Number of Errors", nErrors),
         ReportLongAttribute("n_bytes", "Deleted Bytes", nBytesTotal))
-    val nSkipped = deletedFiles.count(_.nBytes < 0).toLong
+    val nSkipped = deletedFiles.count(_.nBytes < 0)
     if (nSkipped > 0) {
       attributes += ReportLongAttribute("n_skipped", "Skipped Or Missing Files", nSkipped)
     }
--- a/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
+++ b/smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
@@ -5,34 +5,134 @@ import java.nio.file.{Files, Path, Paths}
 import java.io.File
 import java.util.UUID
 
-import org.apache.commons.io.FileUtils
+import org.apache.commons.io.{FileUtils,FilenameUtils}
 import org.joda.time.{DateTime => JodaDateTime}
 
 import scala.util.{Failure, Success, Try}
+import scala.collection.JavaConversions._
+import scala.collection.mutable
 
 import com.pacbio.secondary.analysis.constants.FileTypes
 import com.pacbio.secondary.analysis.reports.ReportUtils
 import com.pacbio.secondary.analysis.reports.ReportModels._
 import com.pacbio.secondary.analysis.tools.timeUtils
 import com.pacbio.secondary.analysis.datasets.io._
+import com.pacbio.secondary.analysis.datasets.DataSetMetaTypes
 import com.pacbio.secondary.analysis.jobs._
 import com.pacbio.secondary.analysis.jobs.JobModels._
+
 import com.pacificbiosciences.pacbiodatasets._
+import com.pacificbiosciences.pacbiobasedatamodel.{InputOutputDataType,ExternalResources}
 
 
-case class DeleteResourcesOptions(path: Path, removeFiles: Boolean = true) extends BaseJobOptions {
+trait DeleteResourcesOptionsBase extends BaseJobOptions {}
+
+case class DeleteResourcesOptions(path: Path, removeFiles: Boolean = true) extends BaseJobOptions with DeleteResourcesOptionsBase {
   def toJob = new DeleteResourcesJob(this)
 }
 
+case class DeleteDatasetOptions(paths: Seq[Path], removeFiles: Boolean = true) extends BaseJobOptions with DeleteResourcesOptionsBase {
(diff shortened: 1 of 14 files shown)


natechols:
@mpkocher any further comments/suggestions before we merge?

smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
```diff
@@ -5,52 +5,67 @@ import java.nio.file.{Files, Path, Paths}
...
+      case Failure(err) => if (ignoreFailures) {
+        logger.error(s"ERROR: ${err.getMessage}"); false
+      } else throw err
     }
+    DeletedFile(f.toString, isDir, size, wasDeleted)
   }
 
-  private def toReport(targetPath: Path, deletedFiles: Seq[DeletedFile]): Report = {
+  protected def toReport(targetPaths: Seq[Path], deletedFiles: Seq[DeletedFile]): Report = {
     val nErrors = deletedFiles.count(_.wasDeleted == false).toLong
```
mpkocher: toLong ?

mpkocher: toLong ?

smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
```diff
@@ -62,64 +77,227 @@ class DeleteResourcesJob(opts: DeleteResourcesOptions)
...
-        ReportStrAttribute("job_dir", "Directory", targetPath.toString),
+    val attributes = mutable.ListBuffer(
+        ReportStrAttribute("job_dir", "Directory", targetPaths.mkString(", ")),
         ReportLongAttribute("n_errors", "Number of Errors", nErrors),
-        ReportLongAttribute("n_bytes", "Deleted Bytes", nBytesTotal)),
-      plotGroups = List[ReportPlotGroup](),
+        ReportLongAttribute("n_bytes", "Deleted Bytes", nBytesTotal))
+    val nSkipped = deletedFiles.count(_.nBytes < 0).toLong
+    if (nSkipped > 0) {
+      attributes += ReportLongAttribute("n_skipped", "Skipped Or Missing Files", nSkipped)
```
mpkocher: Is there a reason `n_skipped` is not always in the report attributes?

Also, if possible, an immutable model would be preferred. 

```scala
val attrs = Seq(1, 2, 3)

val reportAttributes = if (nSkipped > 0 ) attrs else attrs ++ Seq(4)
```

smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
```diff
@@ -62,64 +77,227 @@ class DeleteResourcesJob(opts: DeleteResourcesOptions)
...
+        Seq.empty[DeletedFile]
+      }
+    } else {
+      throw new Exception(s"The path '${jobDir.toString}' does not exist or is not a directory")
+    }
+  }
+}
+
+
+class DeleteDatasetsJob(opts: DeleteDatasetsOptions)
```
mpkocher: Would you mind adding a terse description of the current model as a docstring?

smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
natechols: It's not always there because the DeleteResourcesJob (as currently implemented) will never skip a file - it may fail for some reason but that is counted separately.  In DeleteDatasetsJob, on the other hand, I am making a distinction between delete failures and missing files, the latter being what ends up in n_skipped.  But I'm open to changing this.

smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
```diff
@@ -62,64 +78,226 @@ class DeleteResourcesJob(opts: DeleteResourcesOptions)
...
+ * based on file type, which allows us to exclude any references or barcodes
+ * (etc.) that may be shared with other datasets.
+ */
+class DeleteDatasetsJob(opts: DeleteDatasetsOptions)
+    extends BaseCoreJob(opts: DeleteDatasetsOptions)
+    with DeleteResourcesBase
+    with timeUtils {
+  override val jobTypeId = JobTypeId("delete_dataset")
+  override val resourceType = "Dataset"
+  private val BAM_RESOURCES = Seq(FileTypes.BAM_ALN, FileTypes.BAM_SUB, FileTypes.BAM_CCS, FileTypes.BAM_ALN, FileTypes.I_PBI, FileTypes.I_BAI, FileTypes.STS_XML, FileTypes.STS_H5).map(x => x.fileTypeId).toSet
```
mpkocher: I think renaming this for clarity might help. It's DEFAULT_FILTER_METATYPES or something along this lines.

smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
```diff
@@ -62,64 +78,226 @@ class DeleteResourcesJob(opts: DeleteResourcesOptions)
...
+
+  override def makeReport(files: Seq[DeletedFile]): Report =
+    toReport(opts.paths, files)
+
+  private def getPaths(dsType: DataSetMetaTypes.DataSetMetaType,
+                       externalResources: ExternalResources): Seq[String] = {
+    def isChildResource(e: InputOutputDataType): Boolean = {
+      if (DataSetMetaTypes.BAM_DATASETS contains dsType) {
+        val metaType = e.getMetaType
+        val isChild = (BAM_RESOURCES contains metaType) || metaType.startsWith("PacBio.SubreadFile")
```
mpkocher: This might be useful to abstract into a filter func where getPaths doesn't know anything about filtering mechanism (e: ExternalResource => Boolean). It looks like you're filtering by metatypes from BAM_RESOURCES, but then there's a `metaType.startsWith("PacBio.SubreadFile")` filtering as well.

natechols: The reason for `metaType.startsWith("PacBio.SubreadFile")` is that there are another half-dozen or so metatypes like this that aren't in FileTypes.scala: 

https://github.com/PacificBiosciences/PacBioFileFormats/blob/3.0/DataSet.rst#dataset-external-resource-metatypes-and-file-extensions

But I can add these if you think that would be clearer.

smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobtypes/DeleteResourcesJob.scala
```diff
@@ -62,64 +78,226 @@ class DeleteResourcesJob(opts: DeleteResourcesOptions)
...
+        val dsId = UUID.fromString(ds.getUniqueId)
+        val dsOutPath = s"${dsId}/${dsPath.getFileName.toString}"
+        val dsTmp = Files.createTempFile(s"relativized-${dsId}", ".xml")
+        if (!opts.removeFiles) {
+          logger.info("removeFiles=false, leaving files in place")
+          Seq.empty[DeletedFile]
+        } else {
+          (for {
+            r <- getPaths(dsType, ds.getExternalResources)
+          } yield deleteResource(Paths.get(r))).toList ++
```
mpkocher: Is this `getPaths(dsType, ds.getExternalResources).map(p => deletedResources(Paths.get(p)).toList` the same thing?

natechols:
@skinner can you please double-check my swagger.json changes?

PR #412, opened 2016-12-14
move tools tarball creation to make target

natechols:
@mpkocher any objections?  just trying to clean up some clutter


makefile
```diff
@@ -32,6 +32,15 @@ tools-smrt-analysis-internal:
 tools-smrt-server-analysis:
 	sbt smrt-server-analysis/pack
 
+tools-tarball:
+	sbt smrt-analysis/clean smrt-analysis/compile smrt-analysis/pack
```
mpkocher: Looks good, but maybe try to tighten this up? `sbt clean smrt-server-analysis/pack smrt-analysis/pack`

FYI, multiple commands on a subproject can be called `smrt-server-analysis/{clean, compile,pack}`

There's also this, `sbt pack-archive` however, I don't think that does exactly what we want. 

https://github.com/xerial/sbt-pack

PR #414, opened 2016-12-15
WIP: Fix JobExecutorSpec

smcclellan7:
(Ripped off from @skinner)

### DIFF
--- a/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/services/jobtypes/MockPbsmrtpipeJobType.scala
+++ b/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/services/jobtypes/MockPbsmrtpipeJobType.scala
@@ -7,6 +7,7 @@ import akka.pattern.ask
 import akka.util.Timeout
 import com.pacbio.common.auth.{Authenticator, AuthenticatorProvider}
 import com.pacbio.common.dependency.Singleton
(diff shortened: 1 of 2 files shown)


smcclellan7:
@mpkocher I reverted that commit; I was just using it for debugging.

smcclellan7:
@mpkocher @skinner PTAL

Michael, what you said on slack was correct, the the problem was that the jobs dir is created in production but not in tests, so test jobs can't be completed. See:

https://github.com/PacificBiosciences/smrtflow/blob/master/smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/app/SmrtLinkApp.scala#L60

And when you run the tests locally, as opposed to in circleci, the tests will pass if the jobs dir already exists, which is likely, because the test configs and prod configs use the same path for the jobs dir:

https://github.com/PacificBiosciences/smrtflow/blob/master/smrt-server-link/src/main/resources/reference.conf#L21

https://github.com/PacificBiosciences/smrtflow/blob/master/smrt-server-link/src/test/resources/reference.conf#L11

The change I have hear gets around this problem by creating the jobs dir in the setup block for JobExecutorSpec. That's fine for now, but we might want to generalize that so that it doesn't have to be duplicated in every test that requires jobs to actually complete. (AFAIK, that's currently just the JobExecutorSpec, but this bug was tricky to run down, and I don't want to put anyone else through it, haha.)

PR #415, opened 2016-12-16
services support for soft-delete of datasets

natechols:


mpkocher:
This is a good step forward, however, we probably need a better, consistent model for updating the mutable attributes of the dataset.

PR #418, opened 2016-12-16
logging cleanup

natechols:
Changing the logging of HTTP requests to only display the method, URI, and content (if any).


smrt-server-base/src/main/scala/com/pacbio/common/services/package.scala
```diff
@@ -190,8 +191,16 @@ package object services {
 
     implicit def actorRefFactory = context
 
+    def showRequest(req: HttpRequest): LogEntry = {
+      var asString: String = req.entity.toOption match {
```
mpkocher:
```scala
val asString = req.entity.toOption
   .map(data => s"request: ${req.method} ${req.uri} ${data}")
   .getOrElse(s"request: ${req.method} ${req.uri}")
```

PR #420, opened 2016-12-19
add WhenCompleted to RunDetails and update data model

ben-lerch:


mpkocher:
In the future, I would prefer that Nat, Mitch or myself make these changes. Ideally we only want to checkin the useful diffs of the models that have changed.

PR #421, opened 2016-12-19
[SL-582] re-add job events index, remove runnablejob before callback, bump timeouts

skinner:
This is all related to the timeouts we've been seeing--it brings back the index on job events, fixes the race that allowed jobs to get started more than once, and bumps several timeouts.


skinner:
Closing because George requested a split PR for the index change in SL-582; I'll create two PRs, one for the index change, one for the multiple-job-start and the timeout bumps.

PR #429, opened 2016-12-28
bug fix for pipelines with no task options

natechols:


evolvedmicrobe:
LGTM and we need this to enable HMM fitting, can I go ahead and merge? Last day is tomorrow so it would be great if I could show Kris how to use this then.

PR #435, opened 2017-01-05
extend pipeline template view rules model

natechols:


mpkocher:
We should migrate the view models into a new file, ViewModels or similar.

PR #436, opened 2017-01-05
fix FASTA import/pbsmrtpipe cluster settings bug

natechols:
Also removing the file size cap in `pbservice`.


natechols:
@mpkocher any last-minute thoughts before we merge this?

PR #439, opened 2017-01-10
JSON cleanup and new option types

natechols:


smrt-analysis/src/main/scala/com/pacbio/secondary/analysis/jobs/JobModels.scala
```diff
@@ -57,6 +57,22 @@ object AnalysisJobStates {
 
 }
 
+// mirrors pbcommand.models.common.TaskOptionTypes
+object OptionTypes {
+  sealed trait OptionType {
+    val optionTypeId: String
+  }
+  private def toI(sx: String) = s"pbsmrtpipe.option_types.$sx"
```
mpkocher: There's some comments in the BB pbcommand branch where I mentioned removing the namespacing because of Avro. The new allowed values are not namespaced (e.g., "integer", "choice_integer"). Does this sound reasonable?

mpkocher:
Looks good. I think there might be some issues to shake out (potentially from other components), however, it would probably be best to merge and get this into the nightly build.

PR #442, opened 2017-01-11
export presets as JSON only

natechols:
This is unlovely but it preserves compatibility with existing JSON files (and, more importantly, moves us away from XML).


mpkocher:
Please merge this if the tests pass so we can get this is into the nightly build.

PR #455, opened 2017-01-27
Migrate sqlite indexes

smcclellan7:


smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/database/TableModels.scala
```diff
@@ -176,7 +182,8 @@ object TableModels extends PacBioDateTimeDatabaseFormat {
 
     def name: Rep[String] = column[String]("name")
 
-    def nameUnique = index("project_name_unique", name, unique = true)
+    // Because slick does not support partial indexes, we index this table manually like so:
+    // sqlu"create unique index project_name_unique on projects (name) where is_active;"
```
mpkocher: @smcclellan7 or @skinner Can we simplify this to have the non-partial index? 

Can you clarify how would work in the current model? If I have Project id:1 name:alpha isActive:True and Project id:2 name:alpha isActive:False. Then project 2 is set to active, what is the expected outcome? Does this raise a Constraint error? You can't mark project 2 as active in this case?

skinner: There's currently no undelete option in the services, but if there were, it would result in a constraint error, yes.

```
sqlite> select * from projects;
1|General Project|General Project|CREATED|1461886329576|1461886329576|1
2|foo|foo|CREATED|1|1|1
3|foo|foo|CREATED|1|1|0
sqlite> update projects set is_active=1 where project_id=3;
Error: UNIQUE constraint failed: projects.name
```

I thought it would be confusing if deleted (i.e., not is_active) projects prevented a project of the same name from being created, because you can't see the same-named deleted project that you're colliding with.  But I'm not super attached to that behavior.

A more concrete concern now is if people have deleted projects with potentially-colliding names, and we change the partial index to be non-partial, then the data migration will fail when it hits the colliding names.  Unless we hard-delete projects `where not is_active` as part of the migration, I guess?

Why not just keep it as a partial index?  A while back you were talking about doing migrations in SQL in the postgres era, which I'd prefer (having tried the scala approach).  If we do that, the partial index fits right in with the other SQL.

PR #457, opened 2017-01-27
Update circle.yml to use Postgres 9.6 on 5433

smcclellan7:


mhsieh:
Close this for now, I can help if we need this PR internally.

PR #458, opened 2017-01-30
SL-711 limit project operations by membership role

skinner:
Should this be going into the postgres branch instead of master?  It should be pretty orthogonal to that work, so hopefully it merges relatively cleanly either way.  So let me know if I should rebase and re-point it to a different branch.


mpkocher:
We're currently blocked by DEP to get the integration points working, so please either merge into master if it's a minor change and is absolutely need in the current sprint, then merge master back into the sqlite-to-postgres master. Or if it's not absolutely necessary, then please submit a PR to sqlite-to-postgres branch (the diffs are quite large at this point and it's on the cusp of replacing master when the DEP folks give us the ok).

PR #461, opened 2017-02-01
Rename health service -> alarm service. Create DiskSpaceAlarm.

smcclellan7:


smcclellan7:
@mpkocher - Maybe I'm overthinking this, but with a simpler model, there's a cost in versatility, right?  Like, if we just have a `run()` method that has to compute the metric value from scratch, no context, that works fine for _this particular alarm_, but it doesn't work as well if we want to monitor, say, average latency for API calls, or anything that involves aggregating information over time, or from different sources.  I dunno, maybe there's a way I can simplify, to move in the direction you suggest, without losing all of that flexibility?

mpkocher:
I agree that might be useful, however there as a bit of complexity in your model that was a bit. For example, each alarm had a bit of dependencies, such as the Dao. It also had it's own actor and provider stack. This generated some concern because it required adding a few layers to add a new Alarm to the system. Testing didn't seem that straight-foward as well. 

I believe the Alarm Manager could still enable the functionality that you're trying to preserve while separating concerns. 

The AlarmRunner has Alarm with the core metadata (severity map, window to compute over), but only computes the result a specific point in time. The Alarm Manager can use this Alarm metadata and the computed value (over a window, max, min, mean,...). and update the Dao with the computed Metric? Would this work?

smcclellan7:
@mpkocher what do you think of this model?  Much simplified, much closer to the gist you provided. One difference is that this does not schedule automatic updates, but rather it just updates whenever the value is requested.  I assume the UI will be querying the api at regular intervals to check the alarm status, so there's no point doing extra updates in between requests, or responding to requests without an update in between.  Might as well just let the client set the pace, right?

mpkocher:
I don't think this really makes sense to recompute the alarms current value for every client call. That doesn't sound like it will scale. That will potentially generate a lot of stat calls to the file system.

Nevertheless, I'm going to merge and file a bug to try to fix this in a future sprint.

PR #464, opened 2017-02-03
fix apply-config bug

natechols:
@mpkocher this look right?


mpkocher:
This looks like a bad merge from @smcclellan7 

https://github.com/PacificBiosciences/smrtflow/commit/e0e264d8a80de280f7421e6354225e8c380e847f

PR #467, opened 2017-02-03
Update swagger docs

mpkocher:


mpkocher:
@ben-lerch I've pushed the updated swagger changes here. And deleted the old RST docs that were out of date.

PR #470, opened 2017-02-03
Add support for --version in tools.

mpkocher:


mpkocher:
@natechols Please review. The lack of common interface on some of the *Parser usage introduced some duplication. Those should probably inherit from a common trait to define a concrete interface.

natechols:
I just don't think `scopt` scales very well.  But I'll go for duplication and clarity over inconsistent tools, so LGTM.

PR #474, opened 2017-02-07
Filter data sets by project id

smcclellan7:


smrt-server-link/src/main/scala/com/pacbio/secondary/smrtlink/services/DataSetService.scala
```diff
@@ -84,77 +98,82 @@ class DataSetService(dbActor: ActorRef) extends JobsBaseMicroService with SmrtLi
...
       ma: Marshaller[R],
       sm: Marshaller[Seq[R]]): Route =
-    pathPrefix(shortName) {
-      pathEnd {
-        get {
-          parameter('showAll.?) { showAll =>
-            complete {
-              ok {
-                (dbActor ? GetDataSets(DS_LIMIT, showAll.isDefined)).mapTo[Seq[R]]
+    optionalAuthenticate(authenticator.wso2Auth) { user =>
```
skinner: I thought the optionalAuthenticate had to be inside the HTTP method (get/post/etc.) directive in order for CORSSupport to work.  Am I remembering that correctly?

mpkocher:
It might useful to try to use an immutable model and reuse the query chaining model.

For very common queries, defining q* vals might remove duplication. 

```scala
 val qDataSetMetadataActive = dsMetaData2.filter(_.isActive)

  def getHdfDataSets2(limit: Int = DEFAULT_MAX_DATASET_LIMIT, includeInactive: Boolean = false, projectIds: Seq[Int] = Nil): Future[Seq[HdfSubreadServiceDataSet]] = {
    db.run {
      val qa = if (includeInactive) dsMetaData2 else qDataSetMetadataActive
      val qj = qa join dsHdfSubread2 on (_.id === _.id)
      val q = if (projectIds.isEmpty) qj else qj.filter(_._1.projectId inSet projectIds)
      q.result.map(_.map(x => toHds(x._1, x._2)))
    }
  }
```

smcclellan7:
I tried to change the target branch from sqlite-to-postgres to master, and I broke everything.  It's got all these alarm service changes in it now.  :-P

Should I just make a new branch?

smcclellan7:
@skinner I think you're right about the order of the auth and http method directives in spray, I'll fix that.
@mpkocher An immutable model works fine for me.

skinner:
I think #461 probably should have gone into master, since sqlite-to-postgres had already been merged into master with #407 when #461 was merged.  If we merge sqlite-to-postgres to master again (with the alarm commits from #461), that should make those commits disappear from this PR (since they'll be in both project-filter and master).

The breakage in the latest circle run seems to be from an unresolved merge conflict; I'm not sure where that happened, though.

smcclellan7:
Okay, I created a new branch and cherry-picked the alarm changes into there:

https://github.com/PacificBiosciences/smrtflow/pull/477

Hopefully if we pull that, this PR will look less crazy.

I also fixed the problems with the merge, the order of the optionalAuthorization/get directives, and the mutable slick queries.

skinner:
A commit won't appear in a pull request if the exact same commit (with the same git hash) is present in both the source and target branches.  In that case, there's nothing to pull, because the commit is already there in the destination.

Cherry-picking a commit will (in general) change the hash of the commit, though, because cherry-picking generally changes a commit's parent commit.  So merging a cherry-picked commit in a different PR wouldn't simplify this PR.

At this point, we want to get the alarm changes and these changes into master without having to futz with it too much, right?  But there's something weird going on with the commits on this branch.  Like how 21af40d2 and 44faf40c have different hashes even though they look like they're the same commit.  Maybe something changed in the sqlite-to-postgres->master merge, or maybe something has changed on this branch.

I don't know what's going on, so I'd like to try and simplify the situation.  This is what I think we should do:
  * merge the alarm changes from sqlite-to-postgres into master
  * once that's merged, create a new branch (from master) for the project ID filter work
  * cherry-pick the changes from the project-filter branch into the new branch, making sure there are no merge conflicts
  * create a new PR for the new branch and close this PR.

PR #476, opened 2017-02-07
refactor database tool

natechols:
Plus bonus bug fix to services shutdown script.


smrt-server-analysis/src/main/scala/com/pacbio/secondary/smrtserver/tools/DatabaseTool.scala
```diff
@@ -98,6 +120,19 @@ object DatabaseTool extends CommandLineToolRunner[DatabaseToolOptions] with Engi
...
     db.close()
+  }
+
+  def run(c: DatabaseToolOptions): Either[ToolFailure, ToolSuccess] = {
+
+    val dbConfig = DatabaseConfig(c.dbName, c.username, c.password, c.server, c.port)
+    val startedAt = JodaDateTime.now()
+
+    c.mode match {
+      case DbModes.STATUS => runStatus(dbConfig)
```
mpkocher: If you have `runStatus` return a dbConfig, then you can use function composition.

```scala
val runStatusSummary = runStatus _ andThen runSummary
```

mpkocher:
Looks good.

PR #479, opened 2017-02-08
WIP Add automation constraints

mpkocher:


mpkocher:
This is still WIP, but I would like to unblock the UI. I'm going to add the bundle service, then wire the constraints to be loaded from the bundle.

PR #483, opened 2017-02-08
Add createdBy field to CollectionMetadata

smcclellan7:


mpkocher:
Looks good. @skinner Does this work for you?

skinner:
I don't want to belabor the point, but I just want to note that SqliteToPostgresConverter keeps getting longer.  Aside from that, LGTM.  I assume the Dataset userId -> createdBy model change (and code to populate that from CollectionMetadata) is coming after this?

mpkocher:
@skinner It is indeed going to get longer. At some point we need to copy *all* the model and dependencies into the SqliteToPostgresqConverter to make it completely self contained. I don't see a way around this without rewriting "<>" everywhere (which I can't really motivate). This isn't great, however,  in a future release, we'll completely delete the converter from smrtflow repo.

In the future, please try to provide any input or suggestions before we build out an implementation. Not really sure how much change we can further invest in the converter at the point.

smcclellan7:
@skinner are you referring to SL-853? I can pick that up, unless you've already started working on it?

skinner:
@smcclellan7 I commented in SL-803 with more detail about what I meant.

@mpkocher what do you mean by 'rewriting "<>" everywhere'?
I do provide early input when I can; that's the reason why SqliteToPostgresConverter has the setAutoInc method.  But sometimes people take things in directions that I never would have imagined.

PR #485, opened 2017-02-17
sl-831,859 still trying, not sure if it works

amaster08:


mpkocher:
Please migrate this to BB.

